PR #1, opened 2023-02-11
some review on mate's parser code

Override-6:
some review on mate's parser code, tell me if you like this kind of review, and you can of course review me back if you think some adds are useless or

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -121,7 +121,7 @@ impl<'a> Parser<'a> {
 
     fn next_token(&mut self) -> Result<Token<'a>, ParseError> {
         let token = self.peek_token();
-        if token.token_type == EndOfFile {
+        if token.token_type != EndOfFile {
             self.current += 1;
             Ok(token.clone())
         } else {
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -5,6 +5,7 @@ mod ast;
 
 use crate::ast::*;
 use lexer::token::{Token, TokenType};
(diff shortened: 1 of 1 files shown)


parser/src/lib.rs
```diff
@@ -93,61 +93,55 @@ impl<'a> Parser<'a> {
...
+        self.match_token(expected).ok_or_else(|| self.mk_parse_error(message))
     }
 
-    fn peek_token(&self) -> Token {
-        self.tokens
-            .get(self.current)
-            .cloned()
-            .unwrap_or(Token::new(TokenType::EndOfFile, ""))
+    fn peek_token(&self) -> Token<'a> {
+        self.tokens.get(self.current).cloned().unwrap()
```
syldium, line 119: Calling `.unwrap()` would panic the program if the `Option<Token>` is `None`.

parser/src/lib.rs
```diff
@@ -93,61 +93,55 @@ impl<'a> Parser<'a> {
...
-                true
-            } else {
-                false
-            }
-        } else {
-            false
+    fn exists_token(&mut self, expected: TokenType) -> bool {
+        let token = self.peek_token();
+        if token.token_type == expected {
+            self.current += 1;
```
syldium, line 99: Not a fan of it being named `exists` and actually mutating the `current` token cursor.

syldium (changes requested):

Override-6:
closing this PR as the next PR is based on this one.

PR #2, opened 2023-02-11
parser architecture proposal

Override-6:
Made a first version for the parser architecture. This is to anticipate later features and enhance maintainability. 
here's the parser's file tree :
```
parser/src
├── aspects
│   ├── base_parser.rs
│   ├── call_parser.rs
│   ├── literal_parser.rs
│   └── var_declaration_parser.rs
├── aspects.rs
├── ast.rs
├── lib.rs
└── parser.rs
``` 
We have the `parser.rs` file defining the `ParserResult`, `ParserError` and `Parser` _types_.
The `Parser` type structure is the entrypoint of the whole parser. 
`aspects` are traits (with implementations for the `Parser` structure) for a specific _aspect_ of the parser : function/command call, var declaration, literals etc. (later on, match parser, array parser, block parser, function declaration parser etc)  
the `base_parser` aspect is a special one that contains base functions for parsers

### DIFF
--- a/lexer/src/lexer.rs
+++ b/lexer/src/lexer.rs
@@ -20,4 +20,3 @@ pub fn lex(inp: &str) -> Vec<Token> {
     tokens.push(Token::new(TokenType::EndOfFile, lexer.slice()));
     tokens
 }
-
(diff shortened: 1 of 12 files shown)


syldium (approved):
LGTM

PR #4, opened 2023-02-14
Make the parser aware of space characters

syldium:
The lexer is currently context-free, so cannot differentiate space characters that are in a string literal from the others. The lexer should not be the one that detect string literal because of the existence of string templates that need to be understood by the parser.

The goal of this PR was to maintain having only a single lexer pass. One other goal is to handle filenames as a single identifier. Variable name validation will need to be done after the lexing phase.


parser/src/lib.rs
```diff
@@ -1,32 +1,33 @@
...
-mod parser;
 pub mod ast;
+mod parser;
 
 pub fn parse(tokens: Vec<Token>) -> ParseResult<Vec<Expr>> {
     Parser::new(tokens).parse()
 }
 
 #[cfg(test)]
 mod tests {
```
Override-6, line 18: you created a test folder, why not move all those tests in the folder ?

lexer/src/lexer.rs
```diff
@@ -17,6 +17,99 @@ pub fn lex(inp: &str) -> Vec<Token> {
...
+            vec![
+                Token::new(TokenType::Quote, "'"),
+                Token::new(TokenType::Identifier, "It\\'s"),
+                Token::new(TokenType::Space, " "),
+                Token::new(TokenType::Identifier, "incredible"),
+                Token::new(TokenType::Quote, "'"),
+            ]
+        );
+    }
+}
```
Override-6, line 115: you created a tests folder, why not move those tests in the folder ?

Override-6 (commented):
lgtm excepted some tests location

Override-6 (approved):

PR #12, opened 2023-02-15
Parsing Block expressions, Fixed call and literal parser aspects, Few enhancements in parser cursor

Override-6:
First of all, i apologize for this PR to be a condensation of what could have been several PRs

# Block expressions

Be able to parse code blocks.  
Blocks are essential expressions of the language. they will be used in flow controls, loops, functions bodies etc.

blocks can be empty : 
```scala
{}
{;;;;} //is also an empty block
```
nested : 
```scala
{
   {}
   {}
}
```
but cannot be declared straight after another block expression : 
```scala
{ {} {} } //parse error
```
(needs to be separated with a semicolon (`;`) or a newline) 
another example that currently works : 

```scala
{
   val test = {
      val x = 8
      8
   }
   { val x = 89; command arg1 arg2 }
}
```

# Multiple Calls per lines and escaped literals 
Call parser aspect now able to parse `call1 arg1 arg2 ; call2 arg1 arg2` as two different calls (`call1 arg1 arg2` and `call2 arg1 arg2`). 
If you want to consider the expression as one call with the ';' argument in it, you can escape the semicolon : 
`call1 arg1 arg2 \; call2 arg1 arg2`

new lines can also be escaped : 
```bash
long_call -a -b \
   -c -d -e -f \
   -i 
```

usage example : 
```sh
find -not -type d \
     -iname "*.txt" \
     -exec echo {} \;
```

# Parser Cursor enhancements
Introduced 'or' operation
```rs
///Used to execute `This` or else other if `This` fails
/// returned move is a move that executes either this or other if this move fails.
fn or<B: Move + Clone>(&self, other: B) -> OrMove<This, B>;
```
can now expect a move to be repeated between n and m times :

```rs
///Repeats the given move until it fails, exiting on the first token that made the underlying move fail.
/// NOTE: a repeat always succeed
pub(crate) fn repeat<M: Move + Clone>(mov: M) -> RepeatedMove<M>

///Repeat at least n times the given move until it fails, exiting on the first token that made the underlying move fail.
/// /// if the number of repetition is strictly inferior than n, the move fails
/// NOTE: a repeat always succeed
pub(crate) fn repeat_n<M: Move + Clone>(n: usize, mov: M) -> RepeatedMove<M>

///Repeats between n and m times the given move until it fails, exiting on the first token that made the underlying move fail.
/// if the number of repetition is strictly inferior than n, the move fails
/// if the number of repetition is strictly superior than m, the move also fails
/// NOTE: a repeat always succeed
pub(crate) fn repeat_nm<M: Move + Clone>(n: usize, m: usize, mov: M) -> RepeatedMove<M>
```


parser/src/aspects/block_parser.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn block(&mut self) -> ParseResult<Expr<'a>> {
+        self.expect_token(TokenType::CurlyLeftBracket, "expected start of block expression")?;
+        let block = |exprs| Ok(Expr::Block(Block {
+            exprs
+        }));
+        let mut expressions: Vec<Expr<'a>> = Vec::new();
+        //if the block is empty
+        if self.meet_token(TokenType::CurlyRightBracket) {
+            return block(expressions);
+        }
```
syldium, line 21: Is there any reason to have a spacial case for empty blocks, given that they seem to be handled the same way afterward?

parser/src/aspects/block_parser.rs
```diff
@@ -0,0 +1,194 @@
...
+
+            let expression = self.statement()?;
+            expressions.push(expression);
+            if self.meet_token(TokenType::CurlyRightBracket) {
+                break;
+            }
+
+
+            self.expect_token(TokenType::SemiColon, "")
+                .or_else(|_| self.expect_token(TokenType::NewLine, "expected new line or semicolon"))?;
```
syldium, line 35: The call parser also meets the `TokenType::NewLine`, so it should be removed there.

syldium (commented):
It looks good, there are still some details

parser/src/aspects/block_parser.rs
Override-6: Well, semantically, a function declaration like `fun useless() = {}` would be correct, 
also, having empty blocks in code is useless but still correct. I think we should parse them in the first stage then we'll remove empty blocks when minimizing the AST on later stages

parser/src/aspects/block_parser.rs
Override-6: Not all expressions handle them, Moreover i think any expressions should not handle their termination tokens (newline, semicolons etc) in order to be more flexible with parent expressions such as parenthesis and blocks (i'd be easier to parse childs i think). 
But i think we should reconsider this once the ParserCursor will be OP as its main goal is to facilitate token manipulation.

Override-6: But you right, as the call parser aspect currently consumes the newline, it'll throw here, gotta fix that tomorrow by avoiding the call aspect to consume it.

parser/src/aspects/block_parser.rs
syldium: What I meant is that there is exactly the same condition one line below at the beginning of the loop, so it seems redundant to do the same *is a curly right bracket* check twice. For a first pass, it should indeed not try to minimize the AST.

Override-6: you right, i needed to do this because the method would fail one some cases, let's pause this PR and wait for the ParserCursor to be completed.

Override-6:
closing this PR as the next PR contains it

syldium (changes requested):

parser/src/ast/callable.rs
```diff
@@ -5,6 +5,7 @@ use lexer::token::Token;
 /// A call to a function or a command.
 #[derive(Debug, Clone, PartialEq)]
 pub struct Call<'a> {
+    pub command: Box<Expr<'a>>,
```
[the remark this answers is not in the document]
Override-6: np, i'll rollback that

parser/src/moves.rs
```diff
@@ -15,46 +15,59 @@ pub trait Move {
...
+            right: other,
+        }
+    }
+    fn or<B: Move + Clone>(&self, other: B) -> OrMove<Self, B> {
+        OrMove {
+            left: self.clone(),
+            right: other,
         }
     }
 }
```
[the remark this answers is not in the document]
Override-6: wow ! still need to better understand all that move / borrowing stuff

parser/src/aspects/literal_parser.rs
```diff
@@ -92,26 +92,46 @@ impl<'a> LiteralParser<'a> for Parser<'a> {
...
-            _ => builder.push_str(self.cursor.next()?.value),
+            TokenType::BackSlash => {
+                //if the escaped character is escapable, then we only append the following token's value
+                //else, the backslash is also appended
+                self.cursor.next()?; //advance so we are not pointing to token after '\'
+                if self.cursor.lookahead(escapable()).is_none() {
+                    //if next is not escapable
+                    //will append the backslash '\' char
+                    push_current!();
+                }
```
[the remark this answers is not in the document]
Override-6: ofc but in the actual parser configuration, we'll have to think about this in every parser aspects i think ?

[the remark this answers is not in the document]
Override-6: fixed, thinked bash behaved like that

parser/src/aspects/literal_parser.rs
```diff
@@ -92,26 +92,46 @@ impl<'a> LiteralParser<'a> for Parser<'a> {
...
         loop {
             if self.cursor.is_at_end() {
                 break;
             }
-            match self.cursor.peek().token_type {
-                TokenType::Space => {
-                    self.cursor.advance(next());
+            let pivot = self.cursor.peek().token_type;
+            match pivot {
+                TokenType::Space | TokenType::SemiColon | TokenType::NewLine => {
```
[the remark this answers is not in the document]
Override-6: done

Override-6:
As we are modifying escapement handling, I took the opportunity to make sure that the lexer doesn't handles space escaping anymore.

syldium (approved):
LGTM

PR #14, opened 2023-02-19
introduce Parser cursor 

### LINKED ISSUE #10: Create a ParserCursor structure / trait to navigate in the token stream more easily
a Parser struct would manipulate a `ParserCursor` whose role is to evolve in the token stream, retrieving tokens that matches a predicate, subsequent tokens etc. 

Here's some examples of operations the cursor could do : 
NOTE: the utility functions in `BaseParser` would be moved into the ParserCursor trait
```rs
///ignore all tokens whose types figures in the set, return true if at least one token has been ignored.
ignore(&[TokenType]) -> bool

///ignore all tokens while they satisfy the predicate, return true if at least one token has been ignored.
ignore_while((&Token) -> bool) -> bool

///expect next token type to be present in the set, returning Err(ParseError) otherwise
expect(&[TokenType], Option<&str>) -> ParseResult<Token>

///clone the cursor
clone()
```

Override-6:
# Introduction

Currently, we define our common parsing operations in the `base_parser` aspect. which is a bad practice for several reasons : 

- First, this makes no sense as a parser aspect must parse a unique element of the language (function declaration, call, block, binary operations...). and not be a set of utility methods. 

- Secondly, this procedural paradigm isn't flexible enough, as we cannot do precise parsing operations without defining another method in the `base_parser` aspect, or by spamming `expect`, `peek` and `next` operations which leads in very awful code and harmful maintenance. 

- Thirdly, this is not good codding as we consider utility functions as part of the parser. These functions had to be moved out of the parser.

We need a declarative paradigm in order to do precise and flexible operations. Moreover, a declarative code allows the devs (=parser maintainers) to implement their own "movement" (see below) and then _combine_ it with other movements, creating a more complex parser "movement" route.
In bonus, declarative programming in this case should lead to self-documented code.

Yup, this parser improvement is inspired by [parser combinators](https://en.wikipedia.org/wiki/Parser_combinator).

# The two main structures

## The cursor 
A `ParserCursor` is a mutable structure that contains a vector of token, and it's current position in it.
It's implementation provides the following methods : 
```rs
///advance if next token satisfy the given move
/// Returns the token where the move ended if it succeeds, or None instead.
/// This method will move the current cursor position on where the move ended.
pub fn advance(&mut self, mov: impl Move) -> Option<Token<'a>>

/// Force the given move to succeed, or else fail with given error message.
/// This method will move the current cursor position on where the move ended.
pub fn force(&mut self, mov: impl Move, err: &str) -> ParseResult<Token<'a>>

/// Returns the token where the move ended if the move succeeds, or None instead.
/// This method is similar to `advance` except that it does not makes the cursor change its current pos.
pub fn lookahead(&self, mov: impl Move) -> Option<Token<'a>>
```

All those methods uses a `Move` parameter, which determines the way the cursor will move along its tokens. Those methods only differs on _how_ the cursor will react to the movement. Will it then place its current position at the end of the move ? (`advance/force`) or will it not ? (`lookahead`)

## Movement
`Move` is a single-method trait that inputs an indexed supplier `F` and its position (`A`) in the cursor where the movement has been started. Its role is to output a new position `B` (if succeeded) which will represent the movement (moving from `A` to `B`) in the cursor. 

```rs
trait Move {
    ///F: supplier
    ///usize: A pos (start pos of the movement)
    ///Returns Some(B) if succeed, None instead.
    fn apply<'a, F>(&self, F, usize) -> Option<usize>
    where
        F: Fn(usize) -> Token<'a>;
}
```

There are two main kind of moves : 
- `PredicateMove` which will move cursor to the right (A+1) if its predicate (`Token => bool`) succeeds. input token of the predicate is the token returned by `F(A)`.

- `RepeatedMove` which repeats an underlying move as long as it succeeds.

Here are few builtin moves : 

* `of_type(TokenType...)` -> succeeds if `F(A)` token type matches the given type. 
* `space()` -> match a space
* `no_space()` -> do not match a space
* `next()` -> simply go to the next token.
* `repeat(underlying)` -> repeat the underlying move

### Combination
Moves can be combined using `then` and `and_then` combinations : 
```rs
of_type(Dollar)
   .and_then(of_type(LeftRoundedParenthesis))
   .and_then(of_type(LeftRoundedParenthesis))
``` 
parses `$((`

```rs
of_type(Var)
   .and_then(space().and_then(of_type(Identifier))) //a space between 'var' and the var name is required, this is the diferrence between 'and_then' and 'then'.
   .and_then(space().then(of_type(Colon)))
   .and_then(space().then(of_type(Identifier)))
```
parses a var declaration : `var <identifier name>: <identifier type>`

As you can see, current syntax is quite verbose. Maybe this will be fixed in next updates, but unfortunately, rust does not support implicit, method or operator overloading, so this would remain as is for a long time until someone have a better idea.

closes #10


parser/src/cursor.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn peek(&self) -> Token<'a> {
+        self.at(self.pos)
+    }
+
+    ///returns current token then advance or ParseError if this cursor hits the
+    /// end of the stream.
+    pub fn next(&mut self) -> ParseResult<Token<'a>> {
+        self.next_opt().ok_or(ParseError {
+            message: "Unexpected end of file".to_string(),
+        })
```
syldium, line 64:
```suggestion
        self.next_opt().ok_or_else(|| ParseError {
            message: "Unexpected end of file".to_string(),
        })
```
The string don't need to be allocated every time `next()` is called.

parser/src/aspects/literal_parser.rs
```diff
@@ -133,29 +139,31 @@ impl<'a> LiteralParser<'a> for Parser<'a> {
...
             TokenType::IntLiteral => Ok(LiteralValue::Int(token.value.parse::<i64>().map_err(
-                |e| match e.kind() {
-                    IntErrorKind::PosOverflow | IntErrorKind::NegOverflow => {
-                        self.mk_parse_error("Integer constant is too large.")
+                |e| {
+                    match e.kind() {
+                        IntErrorKind::PosOverflow | IntErrorKind::NegOverflow => self
+                            .expected::<()>("Integer constant is too large.")
+                            .unwrap_err(),
+                        _ => self.expected::<()>(&e.to_string()).unwrap_err(),
```
syldium, line 150: Shouldn't we keep `self.mk_parse_error()` to avoid this `unwrap_err`? `ParseError` can't be initialized directly, as it may depend on the current token in the future.

parser/src/aspects/literal_parser.rs
```diff
@@ -19,42 +19,48 @@ pub(crate) trait LiteralParser<'a> {
...
+                    if token.token_type == TokenType::Quote {
+                        break;
+                    }
+                    value.push_str(token.value);
+                }
+            };
         }
         Ok(Expr::Literal(Literal {
-            token,
+            token: token.clone(),
```
syldium, line 46:
```suggestion
            token,
```
Unnecessary clone

parser/src/aspects/literal_parser.rs
```diff
@@ -114,15 +120,15 @@ impl<'a> LiteralParser<'a> for Parser<'a> {
...
                     }
-                    builder.push_str(self.next_token_space_aware()?.value)
+                    builder.push_str(self.cursor.next()?.value)
                 }
             }
         }
         if !builder.is_empty() {
             parts.push(Expr::Literal(Literal {
-                token: current_start,
+                token: current_start.clone(),
```
syldium, line 131:
```suggestion
                token: current_start,
```
Unnecessary clone

parser/src/cursor.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    ///advance if next token satisfy the given move
+    /// Returns the token where the move ended if it succeeds, or None instead.
+    /// This method will move the current cursor position on where the move ended.
+    pub fn advance(&mut self, mov: impl Move) -> Option<Token<'a>> {
+        let result = mov.apply(|pos| self.at(pos), self.pos);
+
+        if let Some(new_pos) = result {
+            self.pos = new_pos;
+            return Some(self.at(new_pos - 1));
```
syldium, line 30: The `new_pos - 1` seems to contradict the documentation
> Returns the last token before the end of the move if it succeeds?

parser/src/moves.rs
```diff
@@ -0,0 +1,173 @@
...
+    ///Used to chain `This` move with `other` move.
+    /// returns a move that will first execute this move then other one only if this first succeeded.
+    fn and_then<B: Move>(self, other: B) -> AndThenMove<This, B>;
+
+    ///Used to bind `This` move with `other` move.
+    /// returns a move that will first execute this move then the other one.
+    fn then<B: Move>(self, other: B) -> ThenMove<This, B>;
+}
+
+impl<'a, A: Move> MoveOperations<'a, A> for A {
```
syldium, line 29:
```suggestion
impl<A: Move> MoveOperations<A> for A {
```
Unnecessary qualified lifetimes

parser/src/moves.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// `None` if the move did not take effect.
+    ///* `at` - get token at given position
+    ///* `pos` - the position in ParserCursor at beginning of the move
+    fn apply<'a, F>(&self, at: F, pos: usize) -> Option<usize>
+        where
+            F: Fn(usize) -> Token<'a>;
+}
+
+///Defines operations over a Move struct.
+pub(crate) trait MoveOperations<'a, This: Move> {
```
syldium, line 19:
```suggestion
pub(crate) trait MoveOperations<This: Move> {
```
Unnecessary qualified lifetime

parser/src/moves.rs
```diff
@@ -0,0 +1,173 @@
...
+        P: Fn(Token) -> bool,
+{
+    ///The used predicate
+    predicate: P,
+}
+
+impl<P> PredicateMove<P>
+    where
+        P: Fn(Token) -> bool {
+    pub fn negate(self) -> PredicateMove<impl Fn(Token) -> bool> {
```
syldium, line 56:
```suggestion
    /// Invert the current predicate.
    pub fn negate(self) -> PredicateMove<impl Fn(Token) -> bool> {
```

syldium (changes requested):

parser/src/aspects/literal_parser.rs
Override-6: haven't focused on errors as we might rework error handling in a future PR to have clear messages on the relevent error, because current error messages are fatal (one error ends the whole parsing process) and completely unclear to the user (even to us sometimes!) i think we should make a respectable error reporting system in the future, let's discuss that in a later issue.

havn't minded of this, (+ this code comes from you!) i think we could let this as is and rethink all error reports later.

parser/src/cursor.rs
Override-6: It returns the last token that the move operation covered (=validated). The move returns the position on which the cursor must continue. Thus, move operation returns (B), which is the new position of the cursor, but the cursor will return (B-1) as it's the last token covered by the movement. 
I admit that the doc is unclear

PR #15, opened 2023-02-23
Parse command redirections

### LINKED ISSUE #11: Command redirection
A shell allows redirecting file handles, and can change the files the command reads from and writes to. A classical example is redirecting the standard output (file descriptor `1`) of a process:
```kt
ls > /tmp/files // Write the standard output content to a new file /tmp/files
```

By default, the file descriptor is `1`, but the user can work with any open file descriptor:
```kt
example 3> /tmp/output // Write the output of the file handle n°3 to a new file /tmp/output
```

`&>` is a wildcard to redirect both standard output (file descriptor 1) and standard error output (file descriptor 2).

Redirections can be in both directions, so it is common to see:
```kt
command < /tmp/input > /tmp/output // Use the content of /tmp/input as the standard input and write the standard output to /tmp/output
```

A pipeline is a sequence of one or more commands separated by one of the control operators `|`. The format for a pipeline is:
```kt
command1 | command2
```
The standard output of `command1` is connected via a pipe to the standard input of `command2`.

syldium:
The input and output of a command may be redirected using a special notation interpreted by the shell. Redirections can change the files the command reads from and writes to, or duplicate a file descriptor.

Each redirection has a default associated file descriptor, but it may also be preceded by a file descriptor integer literal.
```bash
command < /tmp/input 2> /tmp/output
```

For now, the parser handle:
- Read from a file, write and append redirections to a file. *`> /tmp/out` redirects the standard output to `/tmp/out`.*
- Wildcard redirections. *`nginx &>/dev/null` redirects the standard output and the standard error output to `/dev/null`*
- Duplicate a file descriptor. *`>&2` duplicates the standard output to the standard error output.*
- Command pipelines. *`tr ' ' '\n' < text.txt | sort -u | wc -l` redirects the standard input of `tr`, and link the standard output of this process to the standard input of `sort`. Connect the standard output of `sort` to the standard input of `wc`.*
- Here string. *`grep e <<< 'hello'` feed `hello` to the standard input of `grep`*

Fix #11


parser/src/ast/callable.rs
```diff
@@ -5,8 +5,12 @@ use lexer::token::Token;
 /// A call to a function or a command.
 #[derive(Debug, Clone, PartialEq)]
 pub struct Call<'a> {
-    /// command name at index 0 and its arguments
+    /// The arguments of the command.
+    ///
+    /// A valid command must have at least one argument that is the command name.
     pub arguments: Vec<Expr<'a>>,
+    /// The redirections of the command.
+    pub redirections: Vec<Redir<'a>>,
```
Override-6, line 13: I Think you should move the redirections of a call out of the `Call` structure because it'll be much easier in the future to handle redirections as an `Expr -> Redir` binding instead of a collection of different kinds of redirection for the call. 
Moreover, this current data structure only allows calls to have redirection, but RFC#2 says that this is possible for any kind of expressions, (a block, a match, a loop etc)

parser/src/ast/callable.rs
```diff
@@ -16,3 +20,60 @@ pub struct FunDeclaration<'a> {
...
+
+/// Commands separated by `|`
+#[derive(Clone, Debug, PartialEq)]
+pub struct Pipeline<'a> {
+    /// Elements of the pipeline.
+    ///
+    /// A valid pipeline must have at least one command.
+    pub commands: Vec<Call<'a>>,
+    /// True if the pipeline begins with a `!`.
+    pub negation: bool,
```
Override-6, line 56: I don't think this should be at i the Call structure. As the `!` is an operator, it should be represented somewhere else in the AST imo

Override-6 (commented):
Left some comments on how i think the involved structures should be reviewed in order to anticipate further development steps

Override-6 (changes requested):
oops fail, consider the above comments as part of this request change review

parser/src/aspects/call_parser.rs
```diff
@@ -1,37 +1,223 @@
...
-    use crate::ast::callable::Call;
-    use crate::ast::literal::{Literal, LiteralValue};
-    use crate::ast::Expr;
+    #[test]
+    fn redirection() {
+        let tokens = vec![
+            Token::new(TokenType::Identifier, "ls"),
+            Token::new(TokenType::Greater, ">"),
+            Token::new(TokenType::Identifier, "/tmp/out"),
+        ];
```
Override-6, line 169: For your tests, could you please use the lexer instead of hand-writing the token sequences ? for example in this test, i doubt that the lexer would lexe `/tmp/out` as an identifier. If no, i think it still better to rely on the lexer in order to test the compatibility.

Override-6 (changes requested):
some questions with your tests but otherwise it's good

parser/src/aspects/call_parser.rs
syldium: You're absolutely right. :p

Override-6 (approved):

PR #18, opened 2023-02-26
Use the raw lexme in string literals

### LINKED ISSUE #17: Clearer lexme in the AST
When a string literal is parsed, the parser loses either where the string began or where it ended. This is due to the fact that a literal expression can only have one token in the source string.

To solve this issue, a literal should only have a string slice, the lexme. This would require some unsafe code to concatenate the string slices, but in most cases, the tokens refer to adjacent strings in memory.

syldium:
Some tests that don't use the lexer have been shortened since strings in those tests are not guaranteed to be adjacent in memory.

Fix #17


Override-6 (approved):
Very cool !

PR #19, opened 2023-02-26
Parse arithmetic, comparisons and boolean operations

### LINKED ISSUE #5: Add arithmetic, boolean and comparison operations.
Parse arirhmetic (a + b), boolean (a && b) and comparision (a == b) operations.
This would lead to three new expressions in the AST: 
- `ArithmeticOperation(left: Expr, op: ArithmeticOperator, right: Expr)`
- `BooleanOperation(left: Expr, op: BooleanOperator, right: Expr)`
- `ComparisionOperation(left: Expr, op: BooleanOperator, right: Expr)`

Parser should then be able to parse expressions like `a + b > c && a * b + c == 3.

Override-6:
# Binary operations 
Parser is now able to parse binary operations and handles priority between different operators.

`a + b * c` -> `a + (b * c)`
`(a + b) * c` -> `(a + b) * c`

# Parser context 
Introducing `ParserContext` which is a simple data class that stores information on where the 

Parser is in order to parse some different expressions based on where they are declared : 
```scala
val x = ls //stores 'ls' in x
ls //executes command ls
```

```scala
val a = echo a && echo b //would raise an error
echo a && echo b //echo b if echo a succeeds
```

Other changes
Created a `redirection_parser` aspect that contains all the redirection logic that currently is in `call_parser` aspect

Fixes #5


parser/src/ast/operation.rs
```diff
@@ -1,6 +1,15 @@
...
 use crate::ast::Expr;
+use crate::ast::operation::BinaryOperator::*;
+
+pub const ARITHMETICS: &[BinaryOperator] = &[Plus, Minus, Times, Divide, Modulo];
+pub const COMPARISONS: &[BinaryOperator] = &[EqualEqual, NotEqual, Less, LessEqual, Greater, GreaterEqual];
+pub const BOOLEANS: &[BinaryOperator] = &[And, Or];
+
 
-/// A binary operation between two expressions.
+/// An arithmetic operation between two expressions.
```
syldium, line 12:
```suggestion
/// A binary operation between two expressions.
```
`||` and `&&` are logic operations, not arithmetic.

parser/src/ast/operation.rs
```diff
@@ -11,28 +20,87 @@ pub struct BinaryOperation<'a> {
...
+
+            TokenType::Plus => Ok(Plus),
+            TokenType::Minus => Ok(Minus),
+            TokenType::Star => Ok(Times),
+            TokenType::Slash => Ok(Divide),
+            TokenType::Percent => Ok(Modulo),
+            _ => Err("unexpected non-binary operator token.")
+        }
+    }
 }
```
syldium, line 106: Implement the `TryFrom` trait when defining a fallible way to convert types. Note that this involves further renaming. The documentation is also incorrect, since this code does not panic.
```suggestion
impl TryFrom<TokenType> for BinaryOperator {
    type Error = &'static str;

    /// Convert a TokenType to a BinaryOperator
    fn try_from(token_type: TokenType) -> Result<Self, Self::Error> {
        match token_type {
            TokenType::And => Ok(And),
            TokenType::Or => Ok(Or),

            TokenType::EqualEqual => Ok(EqualEqual),
            TokenType::NotEqual => Ok(NotEqual),
            TokenType::Less => Ok(Less),
            TokenType::LessEqual => Ok(LessEqual),
            TokenType::Greater => Ok(Greater),
            TokenType::GreaterEqual => Ok(GreaterEqual),

            TokenType::Plus => Ok(Plus),
            TokenType::Minus => Ok(Minus),
            TokenType::Star => Ok(Times),
            TokenType::Slash => Ok(Divide),
            TokenType::Percent => Ok(Modulo),
            _ => Err("unexpected non-binary operator token."),
        }
    }
}
```

parser/src/ast/operation.rs
```diff
@@ -1,6 +1,15 @@
+use enum_assoc::Assoc;
+use lexer::token::TokenType;
+
 use crate::ast::Expr;
+use crate::ast::operation::BinaryOperator::*;
+
+pub const ARITHMETICS: &[BinaryOperator] = &[Plus, Minus, Times, Divide, Modulo];
+pub const COMPARISONS: &[BinaryOperator] = &[EqualEqual, NotEqual, Less, LessEqual, Greater, GreaterEqual];
+pub const BOOLEANS: &[BinaryOperator] = &[And, Or];
```
syldium, line 9:
```suggestion
pub const LOGICALS: &[BinaryOperator] = &[And, Or];
```

parser/src/parser.rs
```diff
@@ -70,17 +74,84 @@ impl<'a> Parser<'a> {
...
+        if self.is_at_redirection_sign() {
+            return self.redirectable(left);
+        }
+
+        //if the current token is a binary operator but was not allowed
+        if BinaryOperator::convert_bin_operator(self.cursor.peek().token_type).is_ok() {
+            return self.expected("infix operator not allowed here")
+        }
+        //else a generic error for other kind of tokens.
+        self.expected("invalid token")
```
syldium, line 152: Couldn't the message be clearer?

parser/src/aspects/redirection_parser.rs
```diff
@@ -0,0 +1,267 @@
...
+pub(crate) trait RedirectionParser<'a> {
+    /// Attempts to parse the next pipeline expression
+    /// inputs an "end of call" statements to determine where the call can stop.
+    fn pipeline(&mut self, first_call: Expr<'a>) -> ParseResult<Expr<'a>>;
+    /// Attempts to parse the next redirection
+    fn redirection(&mut self) -> ParseResult<Redir<'a>>;
+    /// Associates any potential redirections to a redirectable expression
+    fn redirectable(&mut self, expr: Expr<'a>) -> ParseResult<Expr<'a>>;
+
+    ///return true if parser is currently on a redirection sign.
```
syldium, line 19:
```suggestion
    /// Tests if the current and subsequent tokens can be part of a redirection expression
```

parser/src/context.rs
```diff
@@ -0,0 +1,60 @@
+use lexer::token::TokenType;
+
+use crate::ast::operation::{ARITHMETICS, BinaryOperator, BOOLEANS, COMPARISONS};
+
+#[derive(Clone)]
+/// A structure that contains contexts
```
syldium, line 6:
```suggestion
/// A structure that contains contexts
#[derive(Clone)]
```

parser/src/context.rs
```diff
@@ -0,0 +1,60 @@
...
+pub struct ParserContext {
+    pub allowed_operators: Vec<BinaryOperator>,
+    pub parsing_value: bool,
+    pub enclosing_end: Option<TokenType>,
+}
+
+impl Default for ParserContext {
+    fn default() -> Self {
+        Self {
+            allowed_operators: [BOOLEANS].concat(),
```
syldium, line 16: Having to allocate a new vector each time a new `ParserContext` is a lot.

parser/src/aspects/redirection_parser.rs
```diff
@@ -0,0 +1,267 @@
...
+                }],
+            })
+        );
+    }
+
+
+    #[test]
+    fn escaped_call() {
+        let tokens = lex("grep -E regex \\; echo test");
+        let parsed = parse(tokens).expect("parsing error");
```
syldium, line 247: Why is this in the redirection parser test? There is no redirection here.

parser/src/aspects/binary_operation_parser.rs
```diff
@@ -0,0 +1,486 @@
...
+                                }),
+                                Expr::Literal(Literal {
+                                    lexme: "world",
+                                    parsed: "world".into(),
+                                }),
+                                Expr::Literal(Literal {
+                                    lexme: "\\)",
+                                    parsed: ")".into(),
+                                }),
+                            ],
```
syldium, line 458:
```suggestion
                        arguments: vec![
                            Expr::Literal("echo".into()),
                            Expr::Literal("hello".into()),
                            Expr::Literal(Literal {
                                lexme: "\\&&",
                                parsed: "&&".into(),
                            }),
                            Expr::Literal("world".into()),
                            Expr::Literal(Literal {
                                lexme: "\\)",
                                parsed: ")".into(),
                            }),
                        ],
```

parser/src/aspects/binary_operation_parser.rs
```diff
@@ -0,0 +1,486 @@
...
+        let right = self.statement(ctx.clone())?;
+
+        //comparison between current operator's priority and next operator (if any)
+        //is 0 if priorities are same or if there is no next operator.
+        //is < 0 if current operator's priority is higher
+        //is > 0 if current operator's priority is smaller
+        let priority_comparison = self.cursor
+            .lookahead(spaces().then(bin_op()))
+            .map(|t| operator_priority - BinaryOperator::convert_bin_operator(t.token_type)
+                .expect("conception error") //cannot fail
```
syldium, line 112:
```suggestion
                .expect("not a valid operator")
```

parser/src/aspects/binary_operation_parser.rs
```diff
@@ -0,0 +1,486 @@
...
+                                priority: i8,
+                                ctx: ParserContext,
+                                left: Expr<'p>,
+    ) -> ParseResult<Expr<'p>> {
+        let mut operation = self.binary_operation_internal(left, ctx.clone())?;
+        macro_rules! hit_eox {
+            () => { self.cursor.lookahead(spaces().then(eox())).is_some() };
+        }
+
+        while !hit_eox!() && self.has_priority(priority) {
```
syldium, line 65: Is the macro really necessary since it is only used once?

syldium (changes requested):
`ParserContext` doesn't seem to be necessary or should be considered as a helper module.
- `parsing_value` is if the parser is in a `statement` or in a `expression`. 
- `allowed_operators` and `with_enclosing_end` should be handled by individuals parsers, in my opinion.

parser/src/aspects/redirection_parser.rs
Override-6: its a fail

parser/src/context.rs
Override-6: i wanted to make constants and statics but it seems not to be possible :(

parser/src/ast/operation.rs
```diff
@@ -1,4 +1,13 @@
+use enum_assoc::Assoc;
+use lexer::token::TokenType;
+
 use crate::ast::Expr;
+use crate::ast::operation::BinaryOperator::*;
+
+pub const ARITHMETICS: &[BinaryOperator] = &[Plus, Minus, Times, Divide, Modulo];
+pub const COMPARISONS: &[BinaryOperator] = &[EqualEqual, NotEqual, Less, LessEqual, Greater, GreaterEqual];
+pub const LOGICALS: &[BinaryOperator] = &[And, Or];
```
syldium, line 9: These constants don't seem to be used anywhere. If they could be used elsewhere, it would be better to have a method on `TokenType` for each group.

parser/src/moves.rs
```diff
@@ -256,25 +272,51 @@ impl<A: Move, B: Move> Move for OrMove<A, B> {
...
+        PredicateMove<for<'a> fn(Token<'a>) -> bool>
     >,
-    PredicateMove<impl (for<'a> Fn(Token<'a>) -> bool)>,
+    PredicateMove<impl ( for<'a> Fn(Token<'a>) -> bool) + Copy>
 > {
-    //if it's escaped then it's not an EOX
-    (of_type(BackSlash).and_then(escapable().negate()))
-        //else it must be either new line or ';'
-        .or(of_types(&[NewLine, SemiColon]))
+    unescaped(of_types(&[SquaredRightBracket, CurlyRightBracket, RoundedRightBracket]))
```
syldium, line 282:
```suggestion
    unescaped(predicate(|t| t.token_type.is_closing_ponctuation()))
```

parser/src/moves.rs
```diff
@@ -256,25 +272,51 @@ impl<A: Move, B: Move> Move for OrMove<A, B> {
 
 //////////////////// STANDARD MOVES ////////////////////
 
-///a move to consume semicolons or new lines as long as they are not escaped.
-pub(crate) fn eox() -> OrMove<
+pub(crate) fn eod() -> OrMove<
```
syldium, line 275: Please document what `eod` means.

parser/src/parser.rs
```diff
@@ -71,13 +155,36 @@ impl<'a> Parser<'a> {
...
 
         while !self.cursor.is_at_end() {
-            statements.push(self.statement()?);
-            self.cursor.advance(space().then(eox()));
+            let statement = self.statement();
+
+            if let Err(error) = &statement {
+                self.report_error(error);
+                self.repos_to_next_expr();
+            }
```
syldium, line 181: Errors should not be printed here, and parsing errors are most likely unrecoverable.

syldium (commented):
This looks great to have all those additions! This will require another review as the changes are quite significant.
The error messages do not need to contain the found token, as it will be underlined by the future error reporter. For now, these messages are fine as they are.

parser/src/ast/operation.rs
Override-6: done

parser/src/parser.rs
```diff
@@ -23,64 +26,105 @@ pub(crate) struct Parser<'a> {
...
 
 impl<'a> Parser<'a> {
+    /// Parses input tokens into an abstract syntax tree representation.
+    pub fn parse(&mut self) -> Vec<Expr<'a>> {
+        let mut statements = Vec::new();
+
+        while !self.cursor.is_at_end() {
+            match self.parse_next() {
+                Err(error) => {
+                    self.report_error(error);
```
syldium, line 36: The parser should not write by itself to the standard error output.

parser/src/parser.rs
```diff
@@ -23,64 +26,105 @@ pub(crate) struct Parser<'a> {
 }
 
 impl<'a> Parser<'a> {
+    /// Parses input tokens into an abstract syntax tree representation.
+    pub fn parse(&mut self) -> Vec<Expr<'a>> {
```
syldium, line 30: The error is lost, keep returning a `ParseResult<Vec<Expr<'a>>` here.

syldium (changes requested):

parser/src/parser.rs
syldium: I've been thinking about this again, certain parse errors are totally recoverable, so having a repos method is totally fine. The first part of the sentence is still valid.

syldium (approved):
Nice work!

PR #21, opened 2023-02-27
Parse command substitution

### LINKED ISSUE #16: Command substitution and evaluation
Any expression that prints to the standard output can be substituted as a string using `$()`.
```kt
val files = $(ls) // Capture stdout of the `ls` process
```

Shells like Bash allows writing `$(<file)` instead of `$(cat file)` to read a string from a file.

The chosen syntax to force expression evaluation (i.e. have a statement in a middle of another) is `@()`.
```kt
val ret = @(ls) // Get the exit status of the `ls` process
```

syldium:
This PR adds support for parsing command substitution.

A substitution is a way to start a new statement in the middle of another.

Command capture:
```kt
val current = $(pwd)
```

Command evaluation:
```kt
val shifted = @(shift)
```

Arithmetic evaluation would wait for #19.

Fix #16

### DIFF
--- a/lexer/src/token.rs
+++ b/lexer/src/token.rs
@@ -152,6 +152,8 @@ impl TokenType {
             TokenType::Ampersand
                 | TokenType::Less
                 | TokenType::Greater
+                | TokenType::And
+                | TokenType::Or
                 | TokenType::Pipe
                 | TokenType::SquareLeftBracket
                 | TokenType::SquareRightBracket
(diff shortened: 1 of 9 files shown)


Override-6 (changes requested):
Would it be possible to add more comments in your implementations please

Override-6 (approved):
lgtm, excepted that it misses some comments on for algorithms

PR #24, opened 2023-03-02
Parse test expressions

### LINKED ISSUE #20: Parse `test` expressions
Use the `[ .. ]` syntax to make a test. 
This test syntax is a test that is evalued by the interpreter.
examples of use : 
```bash
[ ! $a && $b > 78 ] //a binary expression
[ $1 ] //a single-identifier expression
//if no binary or single-identifier expression is detected,
//parse this as a syntactic sugar for the `test` command. Thus, this parses to a `test -d /tmp/f.txt` call
[ -d /tmp/f.txt ] 
```
To force the use of the `test` command, either use `test` directly, or use `[[ .. ]]`
```bash
[[ $a -gt 78 ]] //directly converts to a `test $a -gt 78` call.
```

Override-6:
# Test expression

Adds syntax for test expressions, however, there is a distinction between `[ .. ]` and `[[ .. ]]`
- `[[ .. ]]` is a syntactic sugar to a `test ..` call.
- `[ .. ]` is a test expression directly evaluated by the interpreter.

Thus, those two expressions are drastically different as the `test` command does not have the same syntax as moshell's syntax.

# Not expression
added simple but needed `! x` expression to invert value of `x`.

Fixes #20


syldium (commented):
Nice!

syldium (approved):

PR #26, opened 2023-03-02
Fix #23, enhanced template string lexemes, added `ParserCursor::select` and `NotMove`

### LINKED ISSUE #23: $1, $2 etc throws "Expected variable name"
- `$n` variables with `n` as a name throws parse errors because the parser expects `n` to be an identifier only.
- A `$` alone should not fail for an expectation but be converted to a literal.

Override-6:
- Now able to have variable references like `$1`, `$^`, `$@`, `$!axa!`
only `$$` is allowed though, variables like `$$x` are not 'allowed' and will be parsed as `$$` ref and the `x` will not be part of the expression.

- added `ParserCursor::select` method to return a vector containing all tokens between cursor's initial pos and advanced pos. 

- fixed 'lexme' typo ;)

- added `not` move, look at documentation for further details.

Fixes #23


parser/src/aspects/literal.rs
```diff
@@ -93,30 +93,30 @@ impl<'a> LiteralAspect<'a> for Parser<'a> {
...
+                    } else if let Some(joined) = try_join_str(lexeme, value) {
+                        lexeme = joined;
                     }
                 }
             };
         }
         if !literal_value.is_empty() {
             parts.push(Expr::Literal(Literal {
-                lexme,
+                lexeme: lexeme,
```
syldium, line 119:
```suggestion
                lexeme,
```

parser/src/aspects/literal.rs
```diff
@@ -170,29 +173,29 @@ impl<'a> LiteralAspect<'a> for Parser<'a> {
...
-                    push_current!();
+                    append_current!();
                 }
             }
         }
+
         if !builder.is_empty() {
             parts.push(Expr::Literal(Literal {
-                lexme,
+                lexeme: lexeme,
```
syldium, line 198:
```suggestion
                lexeme,
```

parser/src/aspects/substitution.rs
```diff
@@ -16,14 +17,20 @@ impl<'a> SubstitutionAspect<'a> for Parser<'a> {
...
-            .lookahead(of_type(TokenType::RoundedLeftBracket))
-            .is_none()
+        if self.cursor
+            .lookahead(
+                not(of_type(TokenType::RoundedLeftBracket))
+                    .and_then(
+                        like(TokenType::is_valid_var_ref_name)
+                            .or(of_type(CurlyLeftBracket))
+                            .or(eox())
+                    )
```
syldium, line 31: The look ahead should not look so far. Invalid variable names would not have a meaningful message.

parser/src/ast/variable.rs
```diff
@@ -28,16 +28,16 @@ pub enum VarKind {
...
-    pub name: &'a str,
+    pub name: String,
 }
 
 /// A variable assignation.
 #[derive(Debug, Clone, PartialEq)]
 pub struct Assign<'a> {
     /// The identifier of the variable.
-    pub name: &'a str,
+    pub name: String,
```
syldium, line 40: Variable names do not need to be owned `String`s, especially since user-defined variables have more name restrictions than special variables like `$0` or `$!`.

parser/src/ast/variable.rs
```diff
@@ -28,16 +28,16 @@ pub enum VarKind {
 
 /// A variable reference, prefixed with `$`.
 #[derive(Debug, Clone, PartialEq)]
-pub struct VarReference<'a> {
+pub struct VarReference {
     /// The name of the variable.
-    pub name: &'a str,
+    pub name: String,
```
syldium, line 33: Variable names do not need to be owned `String`s,

syldium (changes requested):

parser/src/aspects/literal.rs
```diff
@@ -93,30 +93,30 @@ impl<'a> LiteralAspect<'a> for Parser<'a> {
                 TokenType::Dollar => {
                     if !literal_value.is_empty() {
                         parts.push(Expr::Literal(Literal {
-                            lexme,
+                            lexeme: lexeme,
```
syldium, line 96:
```suggestion
                            lexeme,
```

parser/src/aspects/substitution.rs
```diff
@@ -14,32 +16,36 @@ pub(crate) trait SubstitutionAspect<'a> {
...
 
         // Short pass for variable references
-        if self
-            .cursor
-            .lookahead(of_type(TokenType::RoundedLeftBracket))
-            .is_none()
-            && start_token.token_type == TokenType::Dollar
+        if self.cursor
+            .lookahead(
+                like(TokenType::is_valid_var_ref_name)
```
syldium, line 38: The lookahead shouldn't check too early that the variable name is valid. This cause that invalid variable names to be ignored by the `var_reference` parser, which would raise a proper and appropriate error for it.

syldium (changes requested):

parser/src/aspects/literal.rs
```diff
@@ -170,29 +173,29 @@ impl<'a> LiteralAspect<'a> for Parser<'a> {
...
-                    push_current!();
+                    append_current!();
                 }
 
-                TokenType::At | TokenType::Dollar => {
+                TokenType::Dollar => {
                     if !builder.is_empty() {
                         parts.push(Expr::Literal(Literal {
-                            lexme,
+                            lexeme: lexeme,
```
syldium, line 182:
```suggestion
                            lexeme,
```

syldium (approved):

PR #34, opened 2023-03-04
Parse if / else statements and expressions

### LINKED ISSUE #28: Parse if / else expressions
Parse `if <expr> { ... } else if { ... } else { ... }` expressions.
if expressions can also be written on one line without using brackets by using a semicolon (`;`) or a newline.

```scala
if [ $1 ] {
   echo "argument $1"
   echo "this is a if expression with a block"
}

if [ $1 ] && echo $1 | grep -E "^[0-9]+$"
   echo $1 is a number ! 
else if [ $1 == "chocolate" ]
   echo "chocolate is good" 
else echo "could not recognize input"

//if / else are expression-statements
val x = if [ {date +"%Y"} < 2023 ]; "bash" else "moshell" 

```

Override-6:
Parse if / else statements, take a look at the issue for further details

Fixes #28


parser/src/parser.rs
```diff
@@ -185,7 +190,8 @@ impl<'a> Parser<'a> {
         self.cursor.advance(spaces()); //consume spaces
 
         //if there is an end of expression, it means that the expr is terminated so we return it here
-        if self.cursor.lookahead(eox().or(eod())).is_some() {
+        //if there is a `else` keyword, maybe that the expression is within a if <expr> else clause
+        if self.cursor.lookahead(eox().or(eod()).or(of_type(Else))).is_some() {
```
syldium, line 194: It might be more readable to add the `Else` token type to the `eod` move. It is strange to have a special case just for `Else`, which is also duplicated below.

parser/src/aspects/if_else.rs
```diff
@@ -0,0 +1,224 @@
...
+        //the success_branch of the if
+        let success_branch = parse_branch(self)?;
+
+        //parse the 'else' branch.
+        let fail_branch =
+            if self.cursor.advance(
+                repeat(spaces().pipe(of_type(NewLine)))
+                    .then(of_type(SemiColon))
+                    .then(repeat(spaces().pipe(of_type(NewLine))))
+                    .then(of_type(Else))
```
syldium, line 41: Couldn't this move be constructed from the previous move, to highlight that only the `.then(of_type(Else))` has been added?

syldium (changes requested):
The pipe move should be renamed: the name *pipe* conveys the idea that we do a move, then another, starting where the last one ended.
It's more like a parallel move.

lexer/src/token.rs
```diff
@@ -148,6 +146,13 @@ pub enum TokenType {
 
 impl TokenType {
 
+    ///is this lexeme a keyword of the language ?
+    pub fn is_keyword(self) -> bool {
+        matches!(self,
+            Fun | Use | IntLiteral | Else | For | In | While | Match | Val | Var
```
syldium, line 152: I don't think that int literals are keywords.

syldium (changes requested):

lexer/src/token.rs
Override-6: fail

syldium (approved):

PR #45, opened 2023-03-07
Add loop / while and "detached" aspects, fix % and / oversights, fix call arguments

### LINKED ISSUE #30: parse while and loop expressions
Introduce `while` and `loop` statements.
`loop` is equivalent to a `while true` expression.
`continue` will go to the next iteration of while / loop 
`break` will break the iterations.

```rs
while [ $1 ] {
   echo "this while is still running"
   match {shift}  {
      -t | -e => ...
      -x | -y => continue 
      _ => break
   }
}
//will never end until connection to root@engine succeeds
loop ssh root@engine && break
```

### LINKED ISSUE #35: call aspect will parse '!' argument as a not expression
doing `echo a ! $b` will parse this as a call to `echo` with arguments `[Literal("a"), Not(VarRef("b"))]` which is not what we want.

`echo a ! $b` should be parsed as `[Literal("a"), Literal("!"), VarRef("b")]`

If we want the negation of `b`, we would do something like `echo a {! b}`

### LINKED ISSUE #39: Allow expressions to be performed in another process
a trailing `&` should execute expression-statements in another proces

```scala
apt install x -y &
while [ $1 ] {
   echo $1
} & //here

match $a {
  ...
} & // here

val x = y & //invalid as val declaration is a statement only

fun x() = {

} & //this would make the function body run in a new process

echo a && b || [ $a == 54 ] && {
   ..
} & //would run all the expression in a new process

echo a; echo b& //would run only echo b

```

### LINKED ISSUE #42: division and modulo operators are not supported (oversight)
we failed to fully implement `/` and `%`

```rs
=> $((4 / 8))
  × invalid expression operator
   ╭─[stdin:1:1]
 1 │ $((4 / 8))
   ·   ┬
   ·   ╰── Here
   ╰────
=> $((4 % 8))
  × invalid expression operator
   ╭─[stdin:1:1]
 1 │ $((4 % 8))
   ·   ┬
   ·   ╰── Here
   ╰────
```

### LINKED ISSUE #44: Value expressions don't handle escaped new lines
As reported while testing #33 
```rs
#[test]
fn arithmetic_multiple_lines() {
    let tokens = lex("val n = 1\\\n + 2");
    let parsed = parse(tokens).expect("Failed to parse");

    assert_eq!(
        parsed,
        vec![Expr::VarDeclaration(VarDeclaration {
            kind: VarKind::Val,
            var: TypedVariable {
                name: "n",
                ty: None,
            },
            initializer: Some(Box::new(Expr::Binary(BinaryOperation {
                left: Box::new(Expr::Literal(Literal {
                    lexeme: "1",
                    parsed: 1.into(),
                })),
                op: BinaryOperator::Plus,
                right: Box::new(Expr::Literal(Literal {
                    lexeme: "2",
                    parsed: 2.into(),
                })),
            }))),
        })],
    );
}
```
`Failed to parse: ParseError { message: "invalid infix operator, found '\\'" }`

Override-6:
Fixes #42
Fixes #39
Fixes #35
Fixes #44
Adds loop / while loop parser (resolves #30)

### DIFF
--- a/lexer/src/token.rs
+++ b/lexer/src/token.rs
@@ -233,8 +233,8 @@ impl TokenType {
     ///is this lexeme a binary operator ?
     pub fn is_bin_operator(self) -> bool {
         match self {
-            And | Or | Plus | Minus | Star | Slash | Percent | EqualEqual | NotEqual | Less | LessEqual | Greater
-            | GreaterEqual => true,
+            And | Or | Plus | Minus | Star | Slash | Percent | EqualEqual | NotEqual | Less
+            | LessEqual | Greater | GreaterEqual => true,
             _ => false,
         }
     }
--- a/lexer/src/token.rs
+++ b/lexer/src/token.rs
@@ -209,6 +209,7 @@ pub enum TokenType {
     #[error]
     Error,
 
+    #[assoc(str = "<end of input>")]
     EndOfFile,
 }
 
@@ -232,7 +233,7 @@ impl TokenType {
     ///is this lexeme a binary operator ?
     pub fn is_bin_operator(self) -> bool {
         match self {
-            And | Or | Plus | Minus | Star | EqualEqual | NotEqual | Less | LessEqual | Greater
+            And | Or | Plus | Minus | Star | Slash | Percent | EqualEqual | NotEqual | Less | LessEqual | Greater
             | GreaterEqual => true,
             _ => false,
         }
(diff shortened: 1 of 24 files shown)


syldium (changes requested):
Please add tests for all the fixed issues.

syldium (approved):
LGTM

PR #49, opened 2023-03-08
Parse structure constructors

### LINKED ISSUE #8: Structure instantiation
Parse structure instantiations, refer to RFC 1.
examples : 
`Ok(value)`
`Array(a b c d)`
`Err("this is an error message")`
Unlick functions and method calls, structure instantiations needs their arguments to get surrounded by parenthesis

syldium:
Fixes #8 by allowing an identifier to be immediately followed by an opening parenthesis. While testing a few edges cases, I also fixed the lexeme for literals, and the escape for strings with simple quotes.


parser/src/parser.rs
```diff
@@ -145,7 +146,14 @@ impl<'a> Parser<'a> {
...
             Match => self.parse_match(Parser::statement).map(Expr::Match),
-
+            Identifier
+                if self
+                    .cursor
+                    .lookahead(of_type(Identifier).and_then(of_type(RoundedLeftBracket)))
+                    .is_some() =>
+            {
+                self.constructor()
+            }
```
Override-6, line 156: isn't it a bit weird that this exact code section also figures in the `Parser::next_value` ? maybe you would like to create your own `newt_x`(let's call that a 'pivot') method,

parser/src/aspects/structure.rs
```diff
@@ -0,0 +1,170 @@
+use crate::ast::structure::Construct;
+use crate::ast::Expr;
+use crate::err::ParseErrorKind;
+use crate::moves::{eod, eox, of_type, word_seps, MoveOperations};
+use crate::parser::{ParseResult, Parser};
+use lexer::token::TokenType;
+
+pub trait StructureAspect<'a> {
+    /// Parses a structure constructor.
```
Override-6, line 9:
```suggestion
    /// Parses a structure constructor call.
```

parser/src/aspects/structure.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        // Read the args until a closing delimiter or a new non-escaped line is found.
+        let mut args = vec![];
+        loop {
+            if self
+                .cursor
+                .advance(word_seps().then(of_type(TokenType::RoundedRightBracket)))
+                .is_some()
+            {
+                return Ok(Expr::Construct(Construct {
```
Override-6, line 32:
```suggestion
self.delimiter_stack.push_back(open_parenthesis.clone());
                return Ok(Expr::Construct(Construct {
```

Override-6: in CLI `Test()` made the parser expect more inputs

parser/src/parser.rs
syldium: Yep totally

parser/src/ast/structure.rs
```diff
@@ -0,0 +1,16 @@
+use crate::ast::Expr;
+use dbg_pls::DebugPls;
+
+/// A constructor call.
+///
+/// This differs from a function call in that it creates a new instance of a
+/// type, rather than perform a specific task. Constructors are always named
+/// and are always called with parentheses.
+#[derive(Debug, Clone, PartialEq, DebugPls)]
+pub struct Construct<'a> {
```
Override-6, line 10: `ConstructorCall` looks more accurate, `Construct` should rather be for Constructor definitions

Override-6 (changes requested):
needs a bit of changes and it'll be ok

parser/src/ast/structure.rs
syldium: *Construct* refers to the action of creating an instance, while *Constructor* refers to the actual definition that is called when constructing an instance. While `ConstructorCall` would work, I think that `Construct` is more in line with the other *short* names we have in the AST.

Override-6 (approved):

PR #59, opened 2023-03-12
Parse for statements

### LINKED ISSUE #29: Parse for in expressions
Parse for expressions, 
`continue` will go to the next iteration of while / loop 
`break` will break the iterations.
```rs
for x in $@ {
   echo $x 
   [ $x == "stop" ] && break;
}
//can also fit in one line 
for x in $@; echo $x 
```

syldium:
There are two kinds of `for` statements. The initializer/condition/increment `for`:

```bash
for ((val i = 0; $i < 10; i=$i + 1)) {
  echo $i
}
```

And the `for` in range loop:
```bash
for n in 1..5; echo $n
for i in (1 + 1)..5; echo $i
for j in $a..$b; echo $j
for arg in $@ {
  echo $arg
}
```

This PR introduces the notion of iterable, which can be a reference to a variable, a file pattern or an int iterator.

Here's an example of a valid file pattern:
```bash
for file in * {
  if test -f $file {
    echo $file
  }
}
```

The parser also provides some interesting parse errors:

```
  × Expected variable name before 'in'
   ╭─[stdin:1:1]
 1 │ for in 0..10; echo $i
   ·     ─┬
   ·      ╰── Here
   ╰────
```
```
  × Receiver variables do not start with '$'
   ╭─[stdin:1:1]
 1 │ for $arg in $@ {
   ·     ┬
   ·     ╰── Here
   ╰────
  help: Consider removing the '$' prefix: for arg in $@
```
And other messages...

Fix #29


parser/src/aspects/loop.rs
```diff
@@ -42,20 +49,239 @@ impl<'a> LoopAspect<'a> for Parser<'a> {
...
+        Ok(RangeFor {
+            receiver: receiver.value,
+            iterable,
+        })
+    }
+
+    /// Parses a "traditional" conditional for, with a initializer, a condition and an increment.
+    fn parse_conditional_for(&mut self) -> ParseResult<ConditionalFor<'a>> {
+        let start = self.cursor.force(
+            repeat_nm(2, 2, of_type(TokenType::RoundedLeftBracket)),
```
Override-6, line 130:
```suggestion
times(2, of_type(TokenType::RoundedLeftBracket)),
```

parser/src/aspects/loop.rs
```diff
@@ -42,20 +49,239 @@ impl<'a> LoopAspect<'a> for Parser<'a> {
...
+            "expected '((' at start of conditional for",
+        )?;
+        for _ in 0..2 {
+            self.delimiter_stack.push_back(start.clone());
+        }
```
Override-6, line 135: why push the _same_ delimiter two times in the stack as a closing delimiter such as `)<x>)` is invalid.

parser/src/aspects/loop.rs
```diff
@@ -42,20 +49,239 @@ impl<'a> LoopAspect<'a> for Parser<'a> {
...
+            if let Expr::VarReference(path) = start {
+                // We don't know yet if the variable is really iterable,
+                // until type checking.
+                return Ok(Iterable::Var(path));
+            }
+            return self.expected_with(
+                "Expected a numeric range, got a single value.",
+                start_token..self.cursor.peek(),
+                Self::suggest_range(&start),
+            );
```
Override-6, line 214: We can't iterate on a predefined item collection :
```rs
for i in a b c d { .. }
// or : 
for i in Array(a b c d) { .. } 
```

parser/src/aspects/loop.rs
```diff
@@ -42,20 +49,239 @@ impl<'a> LoopAspect<'a> for Parser<'a> {
...
+            self.cursor.advance(next());
+            self.var_reference()?
+        } else {
+            self.next_value()?
+        };
+
+        // Test if the range is a single value
+        if self
+            .cursor
+            .advance(repeat_nm(2, 2, of_type(TokenType::Dot)))
```
Override-6:
```suggestion
            .advance(times(2, of_type(TokenType::Dot)))
```

Override-6 (changes requested):

Override-6:
Very good work btw

syldium:
Now implement ranges as values and a subset of file patterns.

Fix #60

Override-6 (approved):

PR #63, opened 2023-03-16
Parse Function Declarations

Override-6:
Functions declarations needs a name and a parameter, an optional return type and type parameter list can also be added.

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -1,6 +1,8 @@
 #![allow(dead_code)]
 #![deny(warnings)]
 
+extern crate core;
+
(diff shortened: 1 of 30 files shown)


parser/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 #![allow(dead_code)]
 #![deny(warnings)]
 
+extern crate core;
+
```
syldium, line 5:
```suggestion
```

syldium (changes requested):

syldium (approved):

PR #68, opened 2023-03-18
Parse function calls with parentheses

### LINKED ISSUE #64: Programmatic Function Call Syntax
We have agreed together that moshell should have two different syntax between a call for an usual command or a simple function, and calls for moshell specific functions, with generic type parameters, and lambda inputs. 

an example of a normal shell call syntax: 

```bash
echo hey, how are you today ?
```

And an example of a PFC syntax
```bash
foo[Str](john, name => echo hey $name this is a lambda !)
```

the call syntax is only forced to be a programatic function call when the parameters of a function are statically known to contain one lambda definition or if generic arguments are explicitely given by the caller (= not infered).

thus, for any other function that does not inputs a lambda, the normal call syntax is useable, as well as the PFC syntax :

```bash
//this
echo this "is a" shell-like "syntax call"
//is equivalent to
echo(this, is a, shell-like, "syntax" call) 
//is also equivalent to
echo(this, "is a", shell-like, syntax call)
```

### LINKED ISSUE #66: Use commas to separate arguments in constructors
in order to align with #64 and keep coherence, we also should force the use of commas to separate arguments for constructors.

syldium:
`Construct` expressions are now `ProgrammaticCall` and separate arguments with a comma.

A `ProgrammaticCall` has its own structure to enforce that the name of the function called is known ahead of time, whereas a regular call may not be.
This also fixes the error message in function declarations when there is no actual parameter between two commas.

~~One noticeable regression is that commas now end raw arguments, meaning that `ls ,,test,` will be split into `ls`, `,`, `,test` and `,`. If we want to keep the old behavior, either raw argument parsing would need to be more context-aware, or programmatic function calls would need to not accept raw arguments.~~

Fix #64
Fix #66


Override-6 (changes requested):
the addition of comma in identifier bounds makes `val x = ,` possible but not `val x = ,,`, maybe we should disallow use of punctuation in this context ? or let them available but then the `,,` case is strange. 
as you have mentioned, the a workaround for this would be to add more context to `LiteralAspect::argument`, i'm not against this idea

syldium:
Commas, arrows and dots are now treated as they are in regular calls only. These tokens still stop raw arguments in the remaining value expressions, since they have an actual meaning in those contexts. `cat comma,separated.txt` reads a single arg for `cat`: `comma,separated.txt`. `var text = comma,separated.txt` is invalid.

Override-6 (changes requested):
`echo 4..5` will parse `4..5` as `4` and `..5`, maybe the range should be parsed also for simpl cal args ?

syldium:
> `echo 4..5` will parse `4..5` as `4` and `..5`, maybe the range should be parsed also for simple call args ?

This PR now comes with another fix: int literals followed by string literals now create only one argument.

Override-6 (approved):

PR #69, opened 2023-03-19
Parse Lambda type declarations

### LINKED ISSUE #65: Lambda type declaration and definition parser
Lambda declaration syntax : 
```scala
A => B
(A, B) => C
(A, Array[B]) => C
```

A lambda that inputs another lambda and returns an output :
```scala
A => B => C
(A => B) => C//same as the parser is left-associative
```

A lambda that returns another lambda :
```scala
A => (A => B)
```

As this is a new kind of type definition,  lambda declaration syntax can be used for any typed thing :
```scala
val x: Int => Str
fun foo(bar: (A, Array[B]) => C)
```

Override-6:
This PR resolves #65 by adding support for lambda type declaration to the parser.
## Type, Simple type and Lambda type
The current parser only supports simple types (a type with a name and optional parameters) which were represented in the AST by the `Type` struct.
This PR introduces a `Type` enum of `Simple` (the current `Type` struct) and `Lambda` (a type that accepts inputs and have an output). 

The syntax to declare a lambda is `In => Out`, where `In` can also be a list of inputs: `(A, B, C) => Out`, `(A) => B` is valid and also `() => B` is valid to declare a lambda with no inputs.

The type enum also contains a Unit variant that represents the Unit type. Two ways of declaring a Unit type is `Unit` o `()`


parser/src/aspects/group.rs
```diff
@@ -141,6 +141,8 @@ mod tests {
     use ast::Expr;
     use context::source::Source;
     use pretty_assertions::assert_eq;
+    use ast::r#type::{Monotype};
```
syldium, line 144:
```suggestion
    use ast::r#type::Monotype;
```

parser/src/aspects/type.rs
```diff
@@ -14,24 +15,119 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
     fn parse_type(&mut self) -> ParseResult<Type<'a>> {
         self.cursor.advance(word_seps()); //consume word seps
 
-        let name_token = self.cursor.next()?;
+
+        let first_token = self.cursor.peek();
+        //if there's a parenthesis then the type is necessarily a polytype
```
syldium, line 20: This is not the case. If there are parentheses, it is a callable type.

parser/src/aspects/type.rs
```diff
@@ -14,24 +15,119 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
...
+
+            if !inputs.is_empty() {
+                rendered_tuple += &format!("{}", inputs.first().unwrap());
+            }
+
+            if inputs.len() > 1 {
+                for tpe in &inputs[1..] {
+                    rendered_tuple += &format!(", {}", tpe);
+                }
+            }
```
syldium, line 82:
```suggestion
            if let Some((first, rest)) = inputs.split_first() {
                write!(rendered_tuple, "{}", first).unwrap();
                for tpe in rest {
                    write!(rendered_tuple, ", {}", tpe).unwrap();
                }
            }
```
Requires importing `use std::fmt::Write;`. It would be worthwhile to write a helper function for that.

parser/src/aspects/type.rs
```diff
@@ -14,24 +15,119 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
...
+
+            rendered_tuple += ") => <type>";
+            return Err(self.mk_parse_error("Tuples are not yet supported. A lambda declaration was expected here",
+                                           self.cursor.peek(),
+                                           Expected(rendered_tuple)))
```
syldium, line 87:
```suggestion
            return self.expected(
                "Tuples are not yet supported. A lambda declaration was expected here",
                Expected(rendered_tuple),
            );
```
The error is also incorrectly positioned.

parser/src/aspects/type.rs
```diff
@@ -14,24 +15,119 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
...
+    fn parse_monotype(&mut self) -> ParseResult<Monotype<'a>> {
+        let name_token = self.cursor.advance(word_seps().then(any())).unwrap();
+
         if name_token.token_type != Identifier {
-            return Err(self.mk_parse_error(
-                format!("'{}' is not a valid type identifier.", name_token.value),
-                name_token,
-                ParseErrorKind::Unexpected,
-            ));
+            return if matches!(name_token.token_type, NewLine | Space) || name_token.token_type.is_closing_ponctuation() {
```
syldium, line 104: This should definitively be a method on `TokenType`.

parser/src/aspects/type.rs
```diff
@@ -14,24 +15,119 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
...
+                Err(self.mk_parse_error(
+                    format!("expected type"),
+                    name_token,
+                    Expected("<type>".to_string()),
+                ))
```
syldium, line 109:
```suggestion
                self.expected_with(
                    "expected type",
                    name_token,
                    Expected("<type>".to_string()),
                )
```

parser/src/aspects/type.rs
```diff
@@ -14,24 +15,119 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
...
+            } else {
+                Err(self.mk_parse_error(
+                    format!("'{}' is not a valid type identifier.", name_token.value),
+                    name_token,
+                    Unexpected,
+                ))
```
syldium, line 115:
```suggestion
                self.expected_with(
                    &format!("'{}' is not a valid type identifier.", name_token.value),
                    name_token,
                    Unexpected,
                )
```

parser/src/aspects/type.rs
```diff
@@ -14,24 +15,119 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
...
+        } else {
+            Type::Monotype(self.parse_monotype()?)
+        };
+
+
+        //check if there's an arrow, if some, we are maybe in a case where the type is "A => ..."
+        // (a lambda with one parameter and no parenthesis for the input, which is valid)
+        if !self.cursor.advance(word_seps().then(of_type(FatArrow))).is_some() {
+            return Ok(tpe)
+        }
```
syldium, line 32:
```suggestion
        if self
            .cursor
            .advance(word_seps().then(of_type(FatArrow)))
            .is_none()
        {
            return Ok(tpe);
        }
```

syldium (changes requested):

parser/src/aspects/type.rs
Override-6: yeah i thought about this, why not introducing a `helper.rs` module in the parser ?

syldium: `helper.rs` is a generic name, I was thinking of having it in the *context* crate instead, along with other functions that generate excerpts from a string or a vector.

ast/src/type.rs
```diff
@@ -1,7 +1,76 @@
...
+#[derive(Debug, Clone, PartialEq, DebugPls)]
+pub struct SimpleType<'a> {
     pub name: &'a str,
     pub params: Vec<Type<'a>>,
 }
+
+#[derive(Debug, Clone, PartialEq, DebugPls)]
+pub struct LambdaType<'a> {
+    pub inputs: Vec<Type<'a>>,
+    pub output: Box<Type<'a>>,
```
syldium, line 24:
```suggestion
    pub params: Vec<Type<'a>>,
    pub returns: Box<Type<'a>>,
```

parser/src/aspects/type.rs
```diff
@@ -183,10 +315,212 @@ mod tests {
...
         );
     }
+
+    #[test]
+    fn lambda_void_type() {
+        // let content = "() => ()";
+        // let source = Source::unknown(content);
+        // let ast = Parser::new(source).parse_type();
+        //todo!()
+    }
```
syldium, line 469:
```suggestion
```

ast/src/type.rs
```diff
@@ -1,7 +1,76 @@
...
+        match self {
+            Type::Simple(m) => Display::fmt(m, f),
+            Type::Lambda(p) => Display::fmt(p, f),
+            Type::Unit => write!(f, "Unit"),
+        }
+    }
+}
+
+///helper function to write a type list format in a given formatter
+fn display_type_list<'a>(
```
syldium, line 38: This function could definitely be moved to a separate file as the code is also shared by https://github.com/moshell-lang/moshell/blob/d91964823796d7255daded79f9767c408d4de5ac/parser/src/aspects/type.rs#L99

ast/src/type.rs
```diff
@@ -1,7 +1,76 @@
...
+use std::fmt::{Debug, Display, Formatter, Write};
+use std::ops::Deref;
 
 #[derive(Debug, Clone, PartialEq, DebugPls)]
-pub struct Type<'a> {
+pub enum Type<'a> {
+    ///A Simple type with optional parameters (`A`, `A[V]`)
+    Simple(SimpleType<'a>),
+    ///A Lambda declaration with of form `(A, B, ...) => Out`
+    Lambda(LambdaType<'a>),
```
syldium, line 10: Wouldn't `CallableType` be more appropriate here? `Lambda` refers to the actual declaration and not the type.

Methods could also be accepted here, for instance:
```kt
fun count(sized: () => Int) -> Int = $sized()
val digits = List[Int](1, 2, 3)
count($digits.size)
```

syldium (changes requested):

ast/src/type.rs
Override-6: no

syldium (approved):
Great work!

PR #71, opened 2023-03-19
Add return keyword

### LINKED ISSUE #67: Return keyword
We forgot to implement the `return` keyword in function declarations !
a `return` always expects a value so the syntaax `return 45 + 81 / 12` needs to be valid

Override-6:
add missing parse of `return` expressions
resolves #67


syldium (approved):
LGTM

PR #74, opened 2023-03-25
Parse lambda declarations

### LINKED ISSUE #73: Parse Lambda argument Declaration

Override-6:
Resolves #73


parser/src/parser.rs
```diff
@@ -202,14 +204,22 @@ impl<'a> Parser<'a> {
...
             Match => self.parse_match(Parser::value).map(Expr::Match),
             Identifier if self.may_be_at_programmatic_call_start() => self.programmatic_call(),
+            Identifier
+                if self
+                    .cursor
+                    .lookahead(any().then(blanks().then(of_type(FatArrow))))
+                    .is_some() =>
+            {
+                self.parse_lambda_definition().map(Expr::LambdaDef)
+            }
```
syldium, line 222: These lines are not covered by any test.
```rs
#[test]
fn lambda_one_arg() {
    let source = Source::unknown("calc(n => $n * $n)");
    let parsed = parse(source).expect("Failed to parse.");
    assert_eq!(
        parsed,
        vec![Expr::ProgrammaticCall(ProgrammaticCall {
            name: "calc",
            arguments: vec![Expr::LambdaDef(LambdaDef {
                args: vec![TypedVariable {
                    name: "n",
                    ty: None,
                }],
                body: Box::new(Expr::Binary(BinaryOperation {
                    left: Box::new(Expr::VarReference(VarReference { name: "n" })),
                    op: BinaryOperator::Times,
                    right: Box::new(Expr::VarReference(VarReference { name: "n" })),
                })),
            })],
            type_parameters: vec![],
        })]
    );
}
```

parser/src/aspects/expr_list.rs
```diff
@@ -0,0 +1,102 @@
...
+        } else {
+            Ok(vec![parse_element(self)?])
+        }
+    }
+
+    fn parse_explicit_list<E, F>(
+        &mut self,
+        start: TokenType,
+        end: TokenType,
+        non_empty: bool,
```
syldium, line 58: I don't think that an "is empty" check should be done by this method, because the error message depends on the method caller.

syldium (changes requested):

parser/src/aspects/expr_list.rs
Override-6: yes indeed

parser/src/parser.rs
Override-6: done

syldium (changes requested):
`echo => 5` is no longer parsed as a classic call, with `["echo", "=>", "5"]` and produces a call to a lambda instead.

syldium (approved):

PR #75, opened 2023-03-26
Initial analyzer work

syldium:
Groundwork for the analyzer.


Override-6:
`ClassType` is a structure that hosts the definition of a type, it defines a type name, their generic parameters (if any), its super type, and the relation between its generics and the generics of the supertype. 
I removed all the `analyzer` stuff as this PR should only define the bases of the type system. I think we may not start to analyse the AST in this first PR and only define how types interacts with each others.
I removed all callable stuffs as this would be too much things to handle for a first Pull Request.

Override-6:
@syldium

PR #83, opened 2023-04-04
Introduction to AST analyzer: datastructure and typesystem early stages

Override-6:
This Pull Request, whose firsts commits were authored by @syldium, lays the bases of the Moshell's source code analyser.
The analyzer's role will be to consume the AST, resolve needed dependencies, perform typechecks, infer types, prevent possible ambiguities and use of unresolved symbols and then output an HIR that'll be later send to the compiler.

This first pull request handles definition of polymorphic types, with generics. It mostly defines the datastructures needed for the typesystem. There is no covariance algorithm, the union algorithm is limited (does not handles type parameters), identity between `TypeClass` does not handles collisions.

There is no way to _start_ the analyzer because this PR is only intended to define a first typesystem and perform typecheck over analyzer's structures. It's up to a later PR to walk the input AST, convert the `ast::Type` to an `analyzer::Type`  and manipulate the typesystem to report errors if any.


analyzer/src/types/class.rs
```diff
@@ -0,0 +1,636 @@
...
+    ///                                  Thus we only have `Map::K => Iterable::A` as an association set.
+    ///
+    /// - `class StrMap[V]: Map[Str, V]` Here, the associations are `lang::Str => Map::K` and `StrMap::V => Map::V`
+    pub(in crate::types) super_params_associations: Vec<ParamAssociation>,
+
+    ///This Type's identity.
+    pub(crate) identity: u64,
+
+    ///The type context of this type.
+    /// The context contains generic parameters definitions
```
syldium, line 40:
```suggestion
    /// The type context of this type.
    ///
    /// The context is guaranteed to own the definition of the generic parameters.
```

analyzer/src/types/class.rs
```diff
@@ -0,0 +1,636 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}: {}", self.name, self.bound)
+    }
+}
+
+pub struct ClassTypeDefinition {
+    super_type: Option<Rc<TypeClass>>,
+    pub name: String,
+    generic_parameters: Vec<GenericParam>,
+    associations: HashMap<usize, Type>,
```
syldium, line 96: It seems wrong to have a `HashMap` indexed like a `Vec`.

ast/src/type.rs
```diff
@@ -13,8 +13,11 @@ pub enum Type<'a> {
     ///A By name declaration (`=> X`)
     ByName(ByName<'a>),
 
-    ///Either `()` or `Unit`, representing a void type
+    ///Either `()` or `Unit`, representing a void types
```
syldium, line 16:
```suggestion
    ///Either `()` or `Unit`, representing a void type
```

parser/src/aspects/type.rs
```diff
@@ -122,7 +119,7 @@ impl<'a> Parser<'a> {
             return self.parse_lambda_with_inputs(inputs).map(Type::Callable);
         }
 
-        //there is no inputs (`()`) and no `=>` after, this is a Unit type
+        //there is no inputs (`()`) and no `=>` after, this is a Unit types
```
syldium, line 122:
```suggestion
        // there is no input (`()`) and no `=>` after, this is a Unit type
```

parser/src/aspects/type.rs
```diff
@@ -23,7 +20,7 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
         self.cursor.advance(blanks());
 
         let first_token = self.cursor.peek();
-        //if there's a parenthesis then the type is necessarily a lambda type
+        //if there's a parenthesis then the type is necessarily a lambda types
```
syldium, line 23:
```suggestion
        // if there's a parenthesis then the type is necessarily a lambda type
```

syldium (changes requested):
This introduction seems epic... :eyes:

analyzer/src/types/context.rs
```diff
@@ -0,0 +1,215 @@
...
+        let defined = Rc::new(defined);
+
+        if ctx.borrow().classes.contains_key(&defined.identity) {
+            return Err(format!("type already contained in context {}", defined.name).to_owned());
+        }
+
+        ctx.borrow_mut()
+            .classes
+            .insert(defined.identity, defined.clone());
+        Ok(defined)
```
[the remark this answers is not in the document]
Override-6: pretty cool :+1:

analyzer/src/types/class.rs
Override-6: This has been done to force the user to explicitly write what index of parent's generic is associated to enhance readability, I'm not a real fan of using a vec here as this would still valid if the changes the parent's type signature

syldium (approved):
Great work!

PR #84, opened 2023-04-05
Parse  `as` expressions for casts

### LINKED ISSUE #79: parse `as` expressions
We need a way to define how to dynamically cast an expression to a certain type.
Example: 
```kt
val x: Any = List("hello", "world") 
x as List[Any]
```

Override-6:
* Resolves #79


parser/tests/expr.rs
```diff
@@ -2,14 +2,15 @@ use ast::call::{Call, ProgrammaticCall, Redir, RedirFd, RedirOp, Redirected};
...
-use ast::r#type::{SimpleType, Type};
+use ast::r#type::{CastedExpr, SimpleType, Type};
 use ast::range::{Iterable, NumericRange};
 use ast::value::{Literal, LiteralValue};
 use ast::variable::{Assign, TypedVariable, VarDeclaration, VarKind, VarReference};
 use ast::Expr;
 use context::source::Source;
 use parser::parse;
 use pretty_assertions::assert_eq;
+use ast::group::{Parenthesis};
```
syldium, line 13:
```suggestion
use ast::group::Parenthesis;
```
This import will most likely be moved up by rustfmt.

parser/src/parser.rs
```diff
@@ -152,31 +153,35 @@ impl<'a> Parser<'a> {
             }
 
             _ => self.next_expression_statement(),
-        }
+        };
+
+        self.handle_cast(expr?)
```
syldium, line 158: Only values can be cast. If this change is reverted, the code still passes all the tests.

syldium (changes requested):

syldium (approved):

PR #85, opened 2023-04-11
Parse method calls

### LINKED ISSUE #72: Parse method calls
```kt
$a.foo(..)
{..}.foo[A](..)
(..).foo(..)
foo(..).bar[A](..)
foo[A](..)(..)
{..}(..)
(..)(..)
$a(..)
```

syldium:
Parse method calls when the dot is immediately followed by an identifier, then by an opening parenthesis/bracket.

```kt
foo().bar[A](9)
create().foo(dummy().truthy()).bar()
$a.foo(1, 2)
```

* Fixes #72


ast/src/call.rs
```diff
@@ -33,8 +33,11 @@ impl SourceSegmentHolder for Call<'_> {
 #[segment_holder]
 #[derive(Debug, Clone, PartialEq, DebugPls)]
 pub struct ProgrammaticCall<'a> {
```
Override-6, line 35: With the introduction of the `MethodCall` structure, you should refactor this to `ProgrammaticFunctionCall`

ast/src/call.rs
```diff
@@ -43,6 +46,25 @@ pub struct ProgrammaticCall<'a> {
...
+#[segment_holder]
+#[derive(Debug, Clone, PartialEq, DebugPls)]
+pub struct MethodCall<'a> {
+    /// The expression on which the method is called.
+    pub source: Box<Expr<'a>>,
+
+    /// The name of the method to call.
+    ///
+    /// The name cannot be an expression, so it is always a constant string after the parsing phase.
+    pub name: &'a str,
```
Override-6, line 59: Use an optional here and do not represent lambda calls (ex: `$a()`) by `ProgrammaticCall`, but use a `MethodCall` with no name instead.

ast/src/call.rs
```diff
@@ -33,8 +33,11 @@ impl SourceSegmentHolder for Call<'_> {
 #[segment_holder]
 #[derive(Debug, Clone, PartialEq, DebugPls)]
 pub struct ProgrammaticCall<'a> {
-    /// The name of the function to call.
-    pub name: &'a str,
+    /// The function to call.
+    ///
+    /// This may or may not be a string literal, as it may a call on a variable, or a call on the
+    /// result of another call.
+    pub expr: Box<Expr<'a>>,
```
Override-6, line 40: You should rollback this to the previous `name: &'a str`, i see you made this to allow expressions like `$a(...)`, `{}(...)` etc to be represented by PFCs but this kind of expression should be represented by `MethodCall` as those are method calls

parser/src/aspects/call.rs
```diff
@@ -26,6 +29,15 @@ pub trait CallAspect<'a> {
     /// Parses a programmatic call.
     fn programmatic_call(&mut self) -> ParseResult<Expr<'a>>;
 
+    /// Parses a programmatic call with a known command name expression.
+    fn programmatic_call_on(&mut self, expr: Expr<'a>) -> ParseResult<Expr<'a>>;
```
Override-6, line 33: This makes no sense to have a _function call_ on an expression (which thus is a method call) if you already defined structures for them

ast/src/call.rs
Override-6, line 40: You should rollback this to the previous `name: &'a str`, i see you made this to allow expressions like `$a(...)`, `{}(...)` etc to be represented by PFCs but this kind of expression should be represented by `MethodCall` as those are method calls

parser/src/aspects/call.rs
Override-6, line 33: This makes no sense to have a _function call_ on an expression (which thus is a method call) if you already defined structures for them

parser/src/aspects/call.rs
```diff
@@ -102,21 +114,72 @@ impl<'a> CallAspect<'a> for Parser<'a> {
...
-            segment: self.cursor.relative_pos(name).start..segment.end,
+            segment: self.cursor.relative_pos(dot).start..segment.end,
         }))
     }
 
+    fn expand_call_chain(&mut self, mut expr: Expr<'a>) -> ParseResult<Expr<'a>> {
+        loop {
+            match self.cursor.peek().token_type {
+                TokenType::RoundedLeftBracket => {
+                    expr = self.programmatic_call_on(expr)?;
```
Override-6, line 167: you should use programmatic_method_on here instead

parser/src/moves.rs
```diff
@@ -165,6 +165,13 @@ pub(crate) fn aerated<M: Move + Copy>(
     blanks().then(m).and_then(blanks())
 }
 
+pub(crate) fn identifier_parenthesis() -> AndThenMove<
+    PredicateMove<impl Fn(Token) -> bool + Copy + Sized>,
+    PredicateMove<impl Fn(Token) -> bool + Copy + Sized>,
+> {
+    of_type(Identifier).and_then(of_types(&[RoundedLeftBracket, SquaredLeftBracket]))
+}
+
```
Override-6, line 174: You should move this under the `STANDARD MOVES` section of the same file
https://github.com/moshell-lang/moshell/pull/85/files#diff-39df3f30535471a9f8520ed353ee96182c28fe313a757245b6082f595608c77fR352

Override-6 (changes requested):
- It seems like you have modified the ProgrammaticCall to support lambda call expressions (ex: `$a.foo(...)(...)`, `{}(...)`. But lambda calls are calls over objects, which makes them a specific kind of method calls with no name. I think that a propper way of representing those method calls is by modifying the MethodCall structure to hold an optional name.

- methods calls on new lines are not suppported : 
```rs
$a.one()
   .dot()
   .per()
   .line()
```
image

Override-6 (approved):

PR #86, opened 2023-04-13
Enhance use expressions and add inclusion path to calls and types

### LINKED ISSUE #78: Parse modules in use statement, pfc and type expressions
```rs
use std::List
my_module::foo[A: std::Option[_]]()
```

### LINKED ISSUE #81: Parse files paths and "@" in uses for environment variables and commands
We need a way to implement other files, and a way to define that we might use an environment variable or an os command.
```rs
use foo/example/script.msh::*
use {@USER, @apt}
```

### LINKED ISSUE #82: Parse stars, type aliases and type lists in use statements
```rs
use std::*
use std::Option
use foo::{Bar, Foo, List}
use bar::foo::{Bar, Foo as MyFoo}

use foo::Bar as foo

```

Override-6:
* Resolves #82 
* Resolves #81
* Resolves #78


context/src/str_find.rs
```diff
@@ -5,7 +5,9 @@ use crate::source::SourceSegment;
 /// # Panics
 /// Panics if the needle is not found.
 pub fn find_in<'a>(source: &'a str, needle: &'a str) -> SourceSegment {
-    let start = source.find(needle).expect("String not found.");
+    let start = source
+        .find(needle)
+        .expect(&format!("'{}' not found.", needle));
```
syldium, line 10:
```suggestion
    let start = match source.find(needle) {
        Some(start) => start,
        None => panic!("'{}' not found.", needle),
    };
```

parser/src/aspects/modules.rs
```diff
@@ -0,0 +1,424 @@
...
+
+            if self
+                .cursor
+                .advance(any().then(spaces().then(of_type(ColonColon))))
+                .is_none()
+            {
+                break;
+            }
+            head.push(identifier)
+        }
```
syldium, line 75:
```suggestion
        while let Some(identifier) = self.cursor.lookahead(spaces().then(of_type(Identifier))) {
            if self
                .cursor
                .advance(any().then(spaces().then(of_type(ColonColon))))
                .is_none()
            {
                break;
            }
            head.push(identifier.value);
        }
```

syldium (changes requested):
Having the base path in a different spot depending on whether you find `use a::a` or `use a::{a}` is very strange.

Override-6:
I added changes in the doc to specify that the path is relative to the expression

syldium (approved):
LGTM

PR #92, opened 2023-04-23
Support local parser errors

### LINKED ISSUE #46: Continue to report errors once one was found in an expression
The current parser is able to report multiple errors per files, but can't report multiple errors per _statements_.

### LINKED ISSUE #89: Multiple single line comments can confuse space handling
Consider the following input:
```kt
{
    //
    //
    ping localhost
}
```

Due to the presence of comments, the lexer will produce the following tokens:
```kt
[
    CurlyLeftBracket,
    NewLine,
    Space, // Before the first comment
    Space, // Before the second comment
    Space, // Before the actual `ping` command
    Identifier,
    Space,
    Identifier,
    NewLine,
    CurlyRightBracket,
]
```

When parsing these tokens, the parser stumbles upon the first argument of `ping` :
```
  × invalid expression operator
   ╭─[../parser/samples/comments.msh:3:1]
 3 │     //
 4 │     ping localhost
   ·          ────┬────
   ·              ╰── Here
 5 │ }
   ╰────
```

One possible fix would be to reintroduce the `repeat_n()` move on the `spaces()` move, removed in #77.
https://github.com/moshell-lang/moshell/blob/4dbb702b0c7184bfedb114fe1aa2150274ee8414/parser/src/moves.rs#L143
```rs
pub(crate) fn spaces() -> RepeatedMove<PredicateMove<impl Fn(Token) -> bool + Copy + Sized>> {
    repeat_n(1, of_type(Space))
}
``

### PR DESCRIPTION
**Pull request by syldium**
Error messages are an important part of Moshell, so this PR goes further by allowing more than one error in a single expression. Some work has been done here to have a smart repositioning to ensure that an error causes the least false positives after. Fix #46 

Repositioning strategy
----------------------

Cursor repositioning depends on two factors :
- The error kind that occurs. If it is a mismatched delimiter error, always advances to the next valid closing delimiter. This is a punctuation error, so it is expected that it could easily go to the end of the file.
- The parser's context. There may sometimes have only one error per line, but if there are clearly defined delimiters like parenthesis or commas, there is no need to go that far.

Repositioning heavily depends on the group which is currently parsed, so it becomes important that those group always keep the delimiter stack in a valid state. In other words, a delimiter scope should always be bound to the appropriate method and should not leak to the caller if there is an error, for instance. It may sound a good use case for a *RAII* object, but that would require a mutable reference to the delimiter stack, which inherently goes against the borrow checker (two mutable references would be required at the same time).

To panic or not to panic
------------------------

Errors were so far returned as `Result<T, ParseError>`. Only one error could be present at a time. This mode can now be called the *panic mode*. When an error has been encountered and the parser don't know how to recover from it, move up the call stack. Some parsers, like list parsers, can handle errors by repositioning to the next `)` or `,` (simplified here). The error is reported - *pushed* - to the error vector shared by the whole parser. Then the parser can continue like nothing happened.

In summary, here are the error phases:

```mermaid
flowchart LR
    raised[Raised]
    panic["Panic mode:\nmove up the call stack"]
    catch[Catched:\nrepositioning and reporting]
    normal[Normal mode]
    raised --> panic
    panic --> catch
    catch --> normal
```

Command Line Interface upgrade
------------------------------

The CLI clearly became a useful tool to quickly test the parser. This PR also updates the CLI to allow parsing an entire file. The REPL style interface is still present and still the default, but adding the `-s` option, handled by the *Clap* crate, makes the CLI reads from a file instead of the standard input.

Existing errors fixed
---------------------

Allowing more than one error reveals some missed cases. Function declarations didn't add the `(` token to the delimiter stack. The group parser failed to parser `{ }` because of the space character. Parsing `use` statements would consume the end of expression character. Fix #89.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -42,8 +42,6 @@ fn main() -> io::Result<()> {
             eprintln!("{msg}");
         }
         exit(1);
-    } else {
-        prompt(handler)?;
     }
-    Ok(())
+    prompt(handler)
 }
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,104 +1,49 @@
+mod cli;
+mod repl;
 mod report;
 
-use crate::report::print_flush;
-use context::source::{Source, SourceSegment};
+use crate::cli::Cli;
+use crate::repl::prompt;
+use crate::report::FormattedError;
+use clap::Parser;
+use context::source::Source;
 use dbg_pls::color;
-use miette::{Diagnostic, GraphicalReportHandler, SourceSpan};
-use parser::err::ParseErrorKind;
+use miette::GraphicalReportHandler;
 use parser::parse;
-use std::fmt::Display;
-use std::io::{self, BufRead, Write};
-use thiserror::Error;
-
-#[derive(Error, Debug, Diagnostic)]
-struct FormattedError<'s> {
-    #[source_code]
-    src: &'s Source<'s>,
-    #[label("Here")]
-    cursor: SourceSpan,
-    #[label("Start")]
-    related: Option<SourceSpan>,
-    message: String,
-    #[help]
-    help: Option<String>,
-}
-
-impl Display for FormattedError<'_> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.message)
-    }
-}
-
-fn offset_empty_span(span: SourceSegment) -> SourceSpan {
-    if span.start == span.end {
-        (span.start - 1..span.end).into()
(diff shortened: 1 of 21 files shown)


cli/src/main.rs
```diff
@@ -1,104 +1,49 @@
...
+        if !msg.is_empty() {
+            eprintln!("{msg}");
+        }
+        exit(1);
+    } else {
+        prompt(handler)?;
     }
-
     Ok(())
 }
```
Override-6, line 49:
```suggestion
        exit(1);
￼    } 
￼    prompt(handler)
￼}
```

Override-6 (approved):

PR #96, opened 2023-05-12
Introduce symbol resolution step

Override-6:
This pull request implements the second step of the checker: the symbol resolution.
While the symbols are collected within the first step, there still some symbols such as an external variable reference that needs to be resolved and linked with the actual variable declaration it's referring to.


analyzer/checker/src/engine.rs
```diff
@@ -36,8 +37,38 @@ impl<'a> Engine<'a> {
...
+    ///Finds an environment by its identifier.
+    pub fn find_environment(&self, id: SourceObjectId) -> Option<&'a Environment> {
+        self.origins.get(id.0).and_then(|(_, env)| {
+            // SAFETY: As the Engine hosts the Environments and may not remove or replace an environment,
+            //The reference behind Environment does not change and is valid for the lifetime of the engine.
+            unsafe {
+                std::mem::transmute::<Option<&Environment>, Option<&'a Environment>>(env.as_ref())
+            }
+        })
     }
```
syldium, line 73:
```suggestion
    pub fn find_environment(&self, id: SourceObjectId) -> Option<&Environment> {
        self.origins.get(id.0).and_then(|(_, env)| {
            env.as_ref()
        })
    }
```

analyzer/checker/src/engine.rs
```diff
@@ -36,8 +37,38 @@ impl<'a> Engine<'a> {
...
+    pub fn attach(&mut self, id: SourceObjectId, env: Environment) -> Result<(), String> {
+        if self.origins[id.0].1.is_some() {
+            return Err(format!(
+                "Could not attach environment to {}: an environment is already attached",
+                id.0
+            ));
+        }
         self.origins[id.0].1 = Some(env);
+        Ok(())
+    }
```
syldium, line 50:
```suggestion
    pub fn attach(&mut self, id: SourceObjectId, env: Environment) {
        debug_assert!(self.origins[id.0].1.is_none(), "Could not attach environment to a source that is already attached");
        self.origins[id.0].1.replace(env);
    }
```
If `attach` is called multiple times for the same object id, it is just a misuse of the API that would not benefit from a `Result`.

analyzer/checker/src/steps/collect.rs
```diff
@@ -0,0 +1,278 @@
...
+                resolver,
+                &mut env,
+                state,
+                expr_idx,
+            )
+            .map_err(GatherError::Other)?;
+        }
+        engine
+            .attach(state.module, env)
+            .map_err(GatherError::Other)?
```
syldium, line 99:
```suggestion
```

analyzer/checker/src/engine.rs
```diff
@@ -36,8 +37,38 @@ impl<'a> Engine<'a> {
...
+    ///Finds an environment by its fully qualified name.
+    pub fn find_environment_by_name(&self, name: &Name) -> Option<SourceObjectId> {
+        for (index, (_, env)) in self.origins.iter().enumerate() {
+            if let Some(env) = env {
+                if &env.fqn == name {
+                    return Some(SourceObjectId(index));
+                }
+            }
+        }
+        None
```
syldium, line 61:
```suggestion
        self.origins
            .iter()
            .position(|(_, env)| {
                if let Some(env) = env {
                    &env.fqn == name
                } else {
                    false
                }
            })
            .map(|index| SourceObjectId(index))
```

syldium (changes requested):

syldium (changes requested):
You may be interested in taking the ownership of the `Resolver::imports` field:
```rs
/// Takes the list of imports that need to be resolved.
pub fn take_imports(&mut self) -> impl Iterator<Item = (SourceObjectId, UnresolvedImports)> {
    std::mem::take(&mut self.imports).into_iter()
}
```
This would be useful to avoid cloning imports and to have a simplified `resolve_symbols` function:
```rs
pub fn resolve_symbols(engine: &Engine, resolver: &mut Resolver) -> Result<(), String> {
    let imports = resolver.take_imports();
    for (env_id, unresolved_imports) in imports {
        let resolved_imports = resolve_imports(engine, unresolved_imports)?;
        let env = engine
            .find_environment(env_id)
            .expect("Could not find environment during resolution");

        resolve_symbols_of(env, resolved_imports, resolver)?
    }

    Ok(())
}
```

syldium (changes requested):

analyzer/system/src/resolver.rs
```diff
@@ -19,6 +20,8 @@ pub struct GlobalObjectId(pub ObjectId);
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub struct SourceObjectId(pub ObjectId);
 
+pub struct ImportsId(pub ObjectId);
+
```
syldium, line 24:
```suggestion
```

syldium (approved):
Great work!

PR #100, opened 2023-05-17
Traverse the whole syntax tree for references

syldium:
Complete syntax tree
-------------------------------

The collection phase now supports every expression type. In order to support functions definitions, environments now hold an optional parent ID. This means that environments may have a lineage, in which variables can be shared. Only global variables should be shared, in order to declare functions in any order. This PR does not implement proper closures support.

Definition annotation
-------------------------------

To prepare error handling and upcoming passes, expressions that refer to a variable are now mapped to their symbol. This mapping uses the internal representation of a `Expr` : the segment is often unique in the AST, but this is not a guarantee. Segments are also bound to a `Source`, meaning that two expressions that share the same segment may actually be quite different. This is a "works for now" solution, if the AST will need more annotations, a new typed AST may simply be created. Annotations are stored in the `Environment` to express that segments should not be used globally. `Environment`s now hold both the expression to variable mapping and the variable ID attribution.

Other changes
---------------------

Empty returns are now valid for the parser.

Local symbols are now always relative from the beginning of the `Vec`, instead of the end, which may change if new variables are created.

Variables are named symbols, which may be function names or actual variables for now.


Override-6:
#100 🥳

Override-6 (approved):

PR #101, opened 2023-05-20
Introduce Diagnostics, with CLI support

Override-6:
This pull request introduces a way for the analyzer to raise diagnostics (either errors or warns).

The collection and resolution steps were improved to be encapsulated into a structure (`SymbolCollector`/`SymbolResolver`}, which lets the methods inside it access a `diagnostics` vector in `self` to push all the diagnostics.
Due to the introduction of the `SymbolResolver`, the current `Resolver` struct has been refactored by `Relations` as it was only a data structure that holds relations between objects/environments.

## The dependency problem and the Importer.
The current `Importer` trait is meant to import a `Source` from a given `Name`, which gives the responsibility to the analyzer to parse the source. This force the analyzer to also handle parse errors, which is not its responsibility: it is only meant to analyze ASTs. 

The `Importer` has been modified to return an AST (`Expr`) from a given `Name`, which moves the responsibility to parse / handle parse errors to the importer's implementation. 

This change makes the `Importer` a bridge between the analyzer and the parser, the CLI implements a simple importer implementation, and is now able to either display parse errors and diagnostics !

_CLI displaying a warning diagnostic raised by the analyzer_
image


cli/src/main.rs
```diff
@@ -1,48 +1,32 @@
...
 use std::process::exit;
 
 fn main() -> io::Result<()> {
     let cli = Cli::parse();
 
-    let handler = GraphicalReportHandler::default();
+    miette::set_hook(Box::new(|_| {
+        Box::new(MietteHandlerOpts::new().tab_width(2).build())
+    }))
+    .expect("miette setup");
```
syldium, line 22:
```suggestion
    miette::set_hook(Box::new(|_| {
        let mut builder = MietteHandlerOpts::new();
        if matches!(env::var("TERM").as_ref().map(String::as_ref), Ok("ansi")) {
            builder = builder.color(true); // Try to detect IntelliJ
        }
        Box::new(builder.build())
    }))
    .expect("miette setup");
```

cli/src/cli.rs
```diff
@@ -9,3 +23,57 @@ pub struct Cli {
...
+
+impl<'a> ASTImporter<'a> for RawImporter<'a> {
+    fn import(&mut self, name: &Name) -> Option<Expr<'a>> {
+        self.imported_modules.get(name).cloned()
+    }
+}
+
+/// Parses and display errors / diagnostics coming from the given source.
+/// Returning true if the source had at least one error or diagnostic.
+pub fn handle_source(source: OwnedSource) -> bool {
```
syldium, line 40: You are not using any special property of `OwnedSource`, use `Source`.

syldium (changes requested):

cli/src/main.rs
Override-6: i'll take a further look on color handlings in [`tree/cli`](https://github.com/moshell-lang/moshell/tree/cli)

cli/src/report.rs
```diff
@@ -43,21 +27,91 @@ fn offset_empty_span(span: SourceSegment) -> SourceSpan {
...
-            },
+pub fn display_parse_error<W: Write>(
+    source: Source,
+    error: ParseError,
+    writer: &mut W,
+) -> io::Result<()> {
+    let span = offset_empty_span(error.position);
+    let mut diag = MietteDiagnostic::new(error.message)
+        .with_severity(Severity::Error)
+        .with_code("error".to_string())
```
syldium, line 38:
```suggestion
```
Parser errors do not have an error code.

context/src/source.rs
```diff
@@ -55,27 +76,7 @@ impl SourceSegmentHolder for Source<'_> {
     }
 }
 
-impl<'s> SourceCode for &'s Source<'_> {
```
syldium, line 58: Why? This seems unrelated to move it in the CLI from the original goal of the PR, simply remove the `&'s` part.

context/src/source.rs
```diff
@@ -7,6 +6,28 @@ pub trait SourceSegmentHolder {
     fn segment(&self) -> SourceSegment;
 }
 
+pub struct StaticSegmentHolder {
```
syldium, line 9: This seems to add a lot of boilerplate. Simply add a constructor to `Observation` that take a raw segment. This type is public, and exposes an API that should be reserved to tests.

cli/src/main.rs
```diff
@@ -1,48 +1,33 @@
+#![allow(dead_code)]
+#![deny(warnings)]
```
syldium, line 2:
```suggestion
```
It's kinda very opinionated to always treat all warnings as errors. It slows the development process when trying to remove code that triggers unused imports, and new warnings may emerge in future versions of Rust. The no warning policy can always be enforced later, for instance in the CI.

syldium (changes requested):

context/src/source.rs
Override-6: The dependency of a diagnostic library such as miette should only be contained by top-level crates. I think that the use of miette in the `context` crate kinda forces the higher crates to depend on it. For example, will we really need to depend on a diagnostic library for a LSP implementation ? ik this is kinda out of scope but i took the advantage of the fact that i was refactoring the cli to fix this useless dependency

syldium: Dependencies are not transitive in Rust, and miette was designed to have its core types in the default crate, and expose the fancy graphical stuff behind a feature flag. I agree that it did break the original "no external dependency" contract, but this concession was made deliberately, in my opinion, to avoid having to create too many `Source`s types that are strictly equivalents. On the other hand, something like `Diagnostic` was completely separated from `ParseError`, as their respective representations differ.

syldium: > For example, will we really need to depend on a diagnostic library for a LSP implementation ?

An editor integration also works with the context, but does not indeed necessary use miette. If this is really a concern, put the implementation behind a feature flag.

> i was refactoring the cli to fix this useless dependency

The dependency is still declared in the `Cargo.toml` file :p

cli/src/repl.rs
```diff
@@ -1,59 +1,49 @@
...
 
-pub fn prompt(handler: GraphicalReportHandler) -> io::Result<()> {
+/// Indefinitely prompts a new expression to the stdin,
+/// displaying back the errors if any and the formed AST
+pub fn prompt() {
+    loop {
+        if let Some(source) = parse_input() {
+            handle_source(source.as_source());
+        }
+    }
```
syldium, line 16:
```suggestion
    while let Some(source) = parse_input() {
        handle_source(source.as_source());
    }
```
The loop must end somewhere where stdin ends.

syldium (approved):
Let's get some fancy import errors :+1:

PR #104, opened 2023-06-04
Support qualified symbols resolution

Override-6:
This pull request introduces the resolution of qualified symbols, which are symbols with an heading path that can either be relative (ex: `io::stdin` where `io` MUST be imported) or absolute (ex: `std::io::stdin`).

## Introduction of collection-resolution Cycles
Currently, the analyzer is straightforward: The first step collects the symbols from an entry point, and follows the links made by `use` statements (the imports) to collect more modules.
Then, once the collection ends, the unresolved symbols are handled by the resolution phase that tries to resolve the relations with other environments variables.

This works, but since this PR introduces qualified symbols, it's now possible to use a symbol from an absolute path, which isn't necessarily in any `use` statements of the current modules set. 
This makes the resolution phase able to request new modules to visit, which then forces the analyzer to alternate between collection and resolution phase as long as new modules are to visit.
The completion of a collection followed by its resolution phase is called a cycle.

This new cycle system requires the resolved imports of environments to persist between cycles.
resolved imports now lives in the `Relations` structure among unresolved imports in a new `Imports` structure

### Introduced `ObjectState`
`Object::resolved: Option<ResolvedSymbol>` has been replaced by `Object::state: ObjectState`.
The state has three variants: `Resolved` (containing the resolved symbol), `Unresolved` and `Dead` which is used to mark objects that raised a diagnostic, and avoids the resolver to attempt to resolve dead objects in other cycles.

## Check symbols names identity
Added a global check after collection phase to ensure that no symbols declared in modules have the same fully-qualified name (= their absolute names) of another module's name. 
This check is required to disambiguate symbols resolution in the resolution step

## Other modifications
- The `Variables::identify` method has been moved into the collection phase as this method is only intended for symbol collectors and the method was doing work that should be let to the collector. Moreover, the method has been enhanced thus it's even more needed to move it in the collector phase.
- `Variables::globals` has been refactored to `Variables::externals` as the map contains resolution requests with external used symbols.
- `Variables::get_reachable` that were returning a `Symbol` now returns an `ObjectID` as it is meant to return local symbols
- `Variables::get_exported` that were returning a `Symbol` now returns an `ObjectID` as it is meant to return external symbols


analyzer/src/steps/resolve.rs
```diff
@@ -101,143 +129,177 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
-            if let Some(resolved_symbol) = imports.imported_symbols.get(symbol_name) {
-                object.resolved = Some(*resolved_symbol);
-            };
+                if env.variables.get_var(var_id).unwrap().is_exported() {
+                    return SymbolResolutionResult::Resolved(symbol);
+                }
+            }
         }
+
+        SymbolResolutionResult::NotFound
```
syldium, line 162:
```suggestion
        if env.has_strict_declaration_order() {
            return SymbolResolutionResult::NotFound;
        }
        if let Some(var_id) = env.variables.get_exported(symbol_name.root()) {
            let symbol = ResolvedSymbol {
                source: env_id,
                object_id: var_id,
            };
            if symbol_name.is_qualified() {
                return SymbolResolutionResult::Invalid(symbol);
            }
            return SymbolResolutionResult::Resolved(symbol);
        }
        SymbolResolutionResult::NotFound
```

analyzer/src/steps/resolve.rs
```diff
@@ -101,143 +129,177 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
+                                imports.set_resolved_import(
+                                    alias.unwrap_or(symbol_name),
+                                    ResolvedImport::Env(found_env_id),
+                                );
+                                continue;
+                            }
                             let symbol_id = found_env
                                 .variables
                                 .exported_vars()
                                 .position(|var| var.name == symbol_name);
```
syldium, line 264:
```suggestion
                            let symbol_id = found_env.variables.get_exported(&symbol_name);
```

analyzer/src/steps/resolve.rs
```diff
@@ -248,36 +310,27 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
-        }
-        None
-    }
+    /// Iterates over remaining unresolved symbols
+    /// This resolution supports parent lookups if the resolution could not
+    fn resolve_trees(&mut self, to_visit: &mut Vec<Name>, visited: &mut HashSet<Name>) {
+        //NOTE: the resolution does not insert new objects
+        'symbol: for object_id in 0..self.relations.objects.len() {
+            let object = &self.relations.objects[object_id];
+            let object_id = GlobalObjectId(object_id);
```
syldium, line 331:
```suggestion
        'symbol: for (object_id, object) in self.relations.objects.iter_mut().enumerate() {
            let object_id = GlobalObjectId(object_id);
```
Maybe move `imports` outside `Resolver` to be able to call `relations.iter_mut()` directly

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +340,219 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
+                    // This happens only if it used it, so we ignore that fact here to always solve external
+                    // symbols via imports.
+                    //
+                    // We omit this resolution for origin environments as the resolution of their own locals
+                    // is done by the collection phase
+                    result = Self::resolve_symbol_from_locals(env_id, env, symbol_name);
+                }
+
+                if result == SymbolResolutionResult::NotFound {
+                    if let Some(imports) = &self.relations.get_imports(&env_id) {
```
syldium, line 366:
```suggestion
                    if let Some(imports) = self.relations.imports.get(&env_id) {
```
Requires to exposes `imports` (or move it outside as mentioned above)

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +340,219 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
                     }
                 }
 
-                // If the symbol is imported, resolve it directly.
-                if let Some(resolved_imports) = resolved_imports.get(&module) {
-                    if let Some(resolved) = resolved_imports.imported_symbols.get(name) {
-                        object.resolved = Some(*resolved);
+                match result {
+                    SymbolResolutionResult::Resolved(symbol) => {
+                        self.relations.objects[object_id.0].state = ObjectState::Resolved(symbol);
```
syldium, line 375:
```suggestion
                        object.state = ObjectState::Resolved(symbol);
```

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +340,219 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
+                            origin_env.find_references(Symbol::Global(object_id.0));
+                        occurrences.sort_by_key(|s| s.start);
+
+                        self.diagnostics.push(diagnose_invalid_symbol(
+                            var.ty,
+                            origin,
+                            symbol_name,
+                            &occurrences,
+                        ));
+                        self.relations.objects[object_id.0].state = ObjectState::Dead;
```
syldium, line 397:
```suggestion
                        object.state = ObjectState::Dead;
```

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +340,219 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
+            } //if the name were already requested, it's definitely unresolvable
+
+            // If we reach this point, the symbol could not be resolved, during any of the previous phases / cycles.
+            self.diagnostics.push(diagnose_unresolved_external_symbols(
+                object_id,
+                origin,
+                origin_env,
+                symbol_name,
+            ));
+            self.relations.objects[object_id.0].state = ObjectState::Dead
```
syldium, line 428:
```suggestion
            object.state = ObjectState::Dead;
```

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +340,219 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
         }
     }
 }
 
+fn get_env_from_relatives<'a>(
+    imports: &Imports,
+    name: &Name,
+    engine: &'a Engine,
+) -> Option<(SourceObjectId, &'a Environment)> {
+    let name_parts = name.parts().to_vec();
```
syldium, line 438:
```suggestion
    let name_parts = name.parts();
```

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +340,219 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
+    segments: &[SourceSegment],
+) -> Diagnostic {
+    let name_root = name.root();
+    let (_, tail) = name.parts().split_first().unwrap();
+    let base_type_name = base_type.to_string();
+    let msg = format!("`{name_root}` is a {base_type_name} which cannot export any inner symbols");
+
+    let observations = segments
+        .iter()
+        .map(|seg| Observation::new((*seg).clone()))
```
syldium, line 489:
```suggestion
        .map(|seg| Observation::new(seg.clone()))
```

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +340,219 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
+        DiagnosticID::UnknownSymbol,
+        env_id,
+        format!("Could not resolve symbol `{name}`."),
+    );
+
+    let mut observations: Vec<_> = env
+        .list_definitions()
+        .filter(|(_, sym)| match sym {
+            Symbol::Local(_) => false,
+            Symbol::Global(g) => g == &relation.0,
```
syldium, line 519:
```suggestion
            Symbol::Global(g) => *g == relation.0,
```

syldium (changes requested):

analyzer/src/steps/resolve.rs
Override-6: the borrowing issue here is because the resolver is mutably borrowed inside the loop when object states are modified

syldium (changes requested):

analyzer/src/environment/variables.rs
```diff
@@ -21,50 +40,46 @@ impl Variables {
...
+
+    /// Returns the local variable associated with the id
+    pub fn find_id(&self, name: &str) -> Option<ObjectId> {
+        self.locals
+            .vars
+            .iter()
+            .rev()
+            .position(|var| var.name == name)
+            .map(|idx| self.locals.vars.len() - 1 - idx)
+    }
```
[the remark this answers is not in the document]
Override-6: it's not equivalent, i was looking for a function that returns any local, regardless if its reachable or exported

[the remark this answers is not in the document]
syldium: Yep, the main interrogation was that this method was not actually used when this comment was written. It's fine now.

analyzer/src/environment/variables.rs
```diff
@@ -214,6 +215,10 @@ impl Variable {
             ty: TypeInfo::Variable,
         }
     }
+
+    pub const fn is_exported(&self) -> bool {
```
syldium, line 219:
```suggestion
    /// Returns `true` if the variable can be accessed externally, without being
    /// captured.
    pub const fn is_exported(&self) -> bool {
```

analyzer/src/name.rs
```diff
@@ -14,7 +14,11 @@ impl Name {
         Self { parts }
     }
 
-    pub fn prefixed(mut path: Vec<String>, name: String) -> Self {
+    pub fn is_qualified(&self) -> bool {
+        self.parts.len() != 1
+    }
+
+    pub fn qualified(mut path: Vec<String>, name: String) -> Self {
```
syldium, line 21:
```suggestion
    /// Creates a new Name from a part and a simple name.
    pub fn qualified(mut path: Vec<String>, name: String) -> Self {
```

analyzer/src/relations.rs
```diff
@@ -122,60 +107,31 @@ pub struct Relations {
...
-        self.objects.push(Object {
-            origin,
-            resolved: None,
-        });
+        self.objects.push(Object::unresolved(origin));
         GlobalObjectId(id)
     }
 
     /// Finds segments that reference the given object.
+    /// Returns non if the object wasn't found or if the tracked object isn't found in the relations
```
syldium, line 121:
```suggestion
    ///
    /// Returns [`None`] if the object is neither found nor tracked.
```

analyzer/src/relations.rs
```diff
@@ -122,60 +107,31 @@ pub struct Relations {
...
         let object = self.objects.get(tracked_object.0)?;
-        let environment = engine.get_environment(object.origin)?;
+        let environment = engine
+            .get_environment(object.origin)
+            .expect("object relation targets to an unknown environment");
         Some(environment.find_references(Symbol::Global(tracked_object.0)))
     }
 
-    /// Returns a mutable iterator over all the objects.
+    /// Returns an immutable iterator over all the objects.
```
syldium, line 134:
```suggestion
    /// Returns a mutable iterator over all the objects.
```

analyzer/src/relations.rs
```diff
@@ -186,7 +142,7 @@ impl Relations {
     /// Returns the resolved symbol for the given object.
     ///
     /// If the object is not resolved or is not referenced, returns `None`.
```
syldium, line 144:
```suggestion
    /// Returns the state of the given object.
    ///
    /// If the object is not referenced, returns [`None`].
```

analyzer/src/steps/resolve.rs
```diff
@@ -75,209 +55,90 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
-    ///
-    /// This resolution should happen after all imports have been resolved in their respective environments,
-    /// to allow child environments to use imports from their parents.
-    fn resolve_trees(&mut self, resolved_imports: &HashMap<SourceObjectId, ResolvedImports>) {
-        'symbol: for (object_id, object) in self.relations.iter_mut() {
-            if object.resolved.is_some() {
+    /// Iterates over remaining unresolved symbols
+    /// This resolution supports parent lookups if the resolution could not
+    fn resolve_trees(&mut self, to_visit: &mut Vec<Name>, visited: &mut HashSet<Name>) {
+        //NOTE: the resolution does not insert new objects
```
syldium, line 139:
```suggestion
```

analyzer/src/steps/resolve.rs
```diff
@@ -75,209 +55,90 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
-
-    /// Iterates over remaining unresolved symbols, and tries to resolve them by traversing the parent chain.
-    ///
-    /// This resolution should happen after all imports have been resolved in their respective environments,
-    /// to allow child environments to use imports from their parents.
-    fn resolve_trees(&mut self, resolved_imports: &HashMap<SourceObjectId, ResolvedImports>) {
-        'symbol: for (object_id, object) in self.relations.iter_mut() {
-            if object.resolved.is_some() {
+    /// Iterates over remaining unresolved symbols
+    /// This resolution supports parent lookups if the resolution could not
```
syldium, line 137:
```suggestion
   /// Iterates over remaining unresolved symbols, and tries to resolve them by traversing the parent chain.
    ///
    /// This resolution should happen after all imports have been resolved in their respective environments,
    /// to allow child environments to use imports from their parents.
```

syldium (changes requested):

analyzer/src/steps/resolve.rs
```diff
@@ -287,53 +149,114 @@ impl<'a, 'e> SymbolResolver<'a, 'e> {
...
+                    let env_var = self
+                        .engine
+                        .get_environment(symbol.source)
+                        .expect("resolved symbol points to an unknown environment");
+                    let var = env_var
+                        .variables
+                        .get_var(symbol.object_id)
+                        .expect("resolved symbol points to an unknown variable in environment");
+                    let mut occurrences: Vec<_> =
+                        origin_env.find_references(Symbol::Global(object_id.0));
```
syldium, line 221:
```suggestion
                    let mut occurrences = origin_env.find_references(object_id.into());
```

analyzer/src/steps/resolve/diagnostics.rs
```diff
@@ -0,0 +1,96 @@
...
+            Symbol::Local(_) => false,
+            Symbol::Global(g) => *g == relation.0,
+        })
+        .map(|(seg, _)| Observation::new(seg.clone()))
+        .collect();
+
+    observations.sort_by_key(|s| s.segment.start);
+    diagnostic.observations = observations;
+
+    diagnostic
```
syldium, line 72:
```suggestion
    let diagnostic = Diagnostic::new(
        DiagnosticID::UnknownSymbol,
        env_id,
        format!("Could not resolve symbol `{name}`."),
    );

    let mut observations: Vec<_> = env
        .list_definitions()
        .filter(|(_, sym)| match sym {
            Symbol::Local(_) => false,
            Symbol::Global(g) => *g == relation.0,
        })
        .map(|(seg, _)| Observation::new(seg.clone()))
        .collect();

    observations.sort_by_key(|s| s.segment.start);
    diagnostic.with_observations(observations)
```
(removed the `mut` modifier and used `with_observations`)

syldium (changes requested):

analyzer/src/steps/resolve.rs
Override-6: i don't like this its not explicit

syldium (approved):

PR #105, opened 2023-06-05
Check type coherence

syldium:
A static typing analysis produces a typed intermediate representation. To bootstrap the process, this is currently done in one pass, that may be extended in the future.

Representing types
----------------------------

Every type has a lot of capacities, but the focus here is its identity. An `Int` is not a `String`. There are the type annotations and the actual types. Type annotation are written by the user and force a specific type. Actual types are deduced from initializers. Those types are associated to the individual variables found during the collection phase. This is a direct mapping, that may be changed for further type inference or to detect if an immutable variable is reassigned.

A minimal set of types is present, but it should be handy to add new ones. `Nothing` will probably be separated into a distinct `Unit` type to better differentiate unreachable code from void returns.

Representing errors
-----------------------------

Checking types may raise a lot of diagnostics, that introduce a type `Error`. The presence of this private type should help for error cascades, it could be associated with any type. Diagnostics are raised as they happen. A few are already implemented, that will hopefully be completed as time advance.

Representing chunks of code
------------------------------------------

Types are added to a new `TypedExpr` type, which can be used to remove unnecessary types at this stage. Function codes are no longer referenced by their parent with a pointer, this is just an ID that matches the existing ID in the existing `Engine`.

A note about exploration planning
--------------------------------------------------

Some environments will always be checked before others. This doesn't matter while the user cannot declare its own types and if function prototypes are not inferred. This latter statement is not always true, so the exploration order matters. This PR tries to get started with ordering by exploring definitions before usages. Due to the way the collection phase unfolds, it is the definition, not the declarations, i.e. the parent environment and not the actual function environment.

In some cases, environments are interdependent, as they share variables or call each other. This must be taken into account during the type ascription phase : we might have a partial knowledge of the other environment, and we cannot directly jump to the other environment to traverse them in parallel. The current implementation work around that by abusing the fact that the site where a function is declared differs from the place where it is actually declared. The strategy is to forward the function declarations to have enough information to work with. This has obviously some drawbacks.
- The parent environment will be typed first, making the order of declarations important.
- Due to the fact that users will depend on the forward declarations because of the collection phase, return type inference cannot be predictable. In other words, it is not guaranteed that users will be explored after the inferred type is inferred.

There are a few possibilities to consolidate this behavior :
- Exposes the signatures of the functions that an environment hoists.
- Do multiple passes to affine the unknown types.


analyzer/src/types/ctx.rs
```diff
@@ -0,0 +1,69 @@
+use crate::relations::{Relations, SourceObjectId, Symbol};
+use crate::types::hir::TypeId;
+use crate::types::{BOOL, FLOAT, INT, NOTHING, STRING};
+use std::collections::HashMap;
+
+pub struct TypeContext {
+    pub(crate) source: SourceObjectId,
```
Override-6, line 7:
```suggestion
    pub(crate) current_source: SourceObjectId,
```
first thought it was a context for each sources :p

analyzer/src/types/ctx.rs
```diff
@@ -0,0 +1,69 @@
...
+            names: HashMap::from([
+                ("Nothing".to_owned(), NOTHING),
+                ("Bool".to_owned(), BOOL),
+                ("Int".to_owned(), INT),
+                ("Float".to_owned(), FLOAT),
+                ("String".to_owned(), STRING),
+            ]),
+            locals: HashMap::new(),
+        }
+    }
```
Override-6, line 25: i think you should rather have a `TypeContext::new` function, and fill-in the lang types elsewhere, this struct is a struct for all the type analysis process and is meant to mutate a lot, thus not really bound to an initial state. also `source: SourceObjectId` is weird, i suppose that it's just here as a default value as the context seems to be _prepared_ before any real use

```suggestion
    pub(crate) fn new(initial_source: SourceObjectId) -> Self {
        let slf = Self {
            source: initial_source,
            names: HashMap::empty(),
            locals: HashMap::new(),
        }
        slf.prepare(initial_source);
        slf
    }
```

analyzer/src/steps/typing.rs
```diff
@@ -0,0 +1,769 @@
...
+        let diagnostics = SymbolCollector::collect_symbols(
+            &mut engine,
+            &mut relations,
+            name.clone(),
+            &mut StaticImporter::new([(name, source)], parse_trusted),
+        );
+        assert_eq!(diagnostics, vec![]);
+        let mut diagnostics = SymbolResolver::resolve_symbols(&engine, &mut relations);
+        assert_eq!(diagnostics, vec![]);
+        let typed = apply_types(&mut engine, &mut relations, &mut diagnostics);
```
Override-6, line 442:
```suggestion
        let typed = apply_types(&engine, &relations, &mut diagnostics);
```

Override-6 (changes requested):

analyzer/src/types/ctx.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl TypeContext {
+    pub(crate) fn new() -> Self {
+        Self {
+            names: HashMap::new(),
+            locals: HashMap::new(),
+        }
+    }
+
+    pub(crate) fn lang() -> Self {
```
Override-6, line 22: this also feels wrong as the context is meant to mutate, i think it'd be better to fill the lang types on a new instance instead of defining a `lang` constructor, or refactor this with `with_lang_prefilled` or something else

analyzer/src/types/ctx.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub(crate) fn resolve(&self, type_annotation: &ast::r#type::Type) -> Option<TypeId> {
+        match type_annotation {
+            ast::r#type::Type::Parametrized(param) => {
+                if !param.path.is_empty() || !param.params.is_empty() {
+                    return None;
+                }
+                self.names.get(param.name).copied()
+            }
+            ast::r#type::Type::Callable(_) => None,
+            ast::r#type::Type::ByName(_) => None,
```
Override-6, line 77:
```suggestion
                    unimplemented!();
                    //return
                }
                self.names.get(param.name).copied()
            }
            ast::r#type::Type::Callable(_) => unimplemented!(),
            ast::r#type::Type::ByName(_) => unimplemented!(),
```

analyzer/src/types/hir.rs
```diff
@@ -0,0 +1,74 @@
...
+/// An expression content.
+#[derive(Debug, PartialEq)]
+pub enum ExprKind {
+    Literal(LiteralValue),
+    Assign {
+        lhs: Box<TypedExpr>,
+        rhs: Box<TypedExpr>,
+    },
+    Declare {
+        identifier: Symbol,
```
Override-6, line 52: we can declare external symbols ? wouldn't it just be an `ObjectId` or a `ResolvedSymbol`, i don't see cases where we _declare_ a variable that is also declared elsewhere, or maybe i'm missing something

Override-6 (changes requested):

analyzer/src/relations.rs
```diff
@@ -189,4 +190,19 @@ impl Relations {
     pub fn get_resolved(&self, id: GlobalObjectId) -> Option<ResolvedSymbol> {
         self.objects.get(id.0)?.resolved
     }
+
+    /// Creates a dependency graph for the given engine.
+    pub fn build_dependencies(&self, engine: &Engine) -> Dependencies<SourceObjectId> {
```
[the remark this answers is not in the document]
syldium: This method is placed in `Relations`, so it builds a relation graph. `as_graph()` might be a better name if we completely ignore the context around this method.

analyzer/src/types/ctx.rs
syldium: It is a default constructor, that explicitly states that it initially contains the definition for primitive types. Creating a really empty `TypeContext` is something that should be disallowed, since most of the consumers expect the primitives types to always be present. See the constants that always points to the same index.
This context will be mutated, but already inserted types will never be removed. Merging this constructor with the `fill_lang` method would be the best move in my opinion.

analyzer/src/types/hir.rs
syldium: Yep, you're right!

analyzer/src/types/ctx.rs
Override-6: the definition of `TypeContext#lang` in this way lets think that the context is bound to the lang types, the primitives. If i mutate it (i.e push a new type into it), it can feels like i'm creating a new lang type, which isn't, i agree with the fact that we cannot create a type context without the lang types, thus indeed hiding the `fill_lang` inside `new` or creating a `with_lang_prefilled` can be a good idea

Override-6 (approved):

PR #109, opened 2023-06-15
Refactor VM, Compiler and Typing state, add jump support to VM and while/loop to compiler

Override-6:
This pull request adds loop and while support to the compiler, and adds three new JUMP operands:
- `OP_JUMP` to jump to the given instruction pointer
- `OP_IF_JUMP` to jump to the given instruction pointer if the last value in operand stack is equal to 0
- `OP_IF_NOT_JUMP` to jump to the given instruction pointer if the last value in operand stack is not equal to 0

Also needed to handle value pop more specifically, thus two new POP instruction were introduced to replace the old `OP_POP`:
`OP_POP_BYTE` to pop one byte from the operand stack (for exitcodes, bool)
`OP_POP_Q_WORD` to pop 8 bytes from the operand stack (for string references, int, float)


analyzer/src/steps/typing.rs
```diff
@@ -122,339 +172,521 @@ fn apply_types_to_source(
...
+        kind: ExprKind::ConditionalLoop(Loop {
+            condition: condition.map(Box::new),
+            body: Box::new(body),
+        }),
+        segment: loo.segment(),
+        ty: NOTHING,
+    }
+}
+
+fn ascribe_continue_or_break(
```
syldium, line 601: This function is not tested.

compiler/src/bytecode.rs
```diff
@@ -19,23 +36,51 @@ impl Bytecode {
...
+    }
+
+    pub fn extend(&mut self, bytecode: Bytecode) {
+        self.bytes.extend(bytecode.bytes);
+    }
+
+    /// Fills an instruction pointer at given instruction pointer in the byte array
+    pub fn fill_in_ip(&mut self, ip_dest: usize, ip: usize) {
+        self.bytes[ip_dest..ip_dest + size_of::<usize>()]
+            .copy_from_slice(ip.to_be_bytes()[..].try_into().unwrap())
```
syldium, line 63:
```suggestion
        self.bytes[ip_dest..ip_dest + size_of::<usize>()].copy_from_slice(&ip.to_be_bytes())
```

analyzer/src/steps/typing.rs
```diff
@@ -122,339 +172,521 @@ fn apply_types_to_source(
     }
 }
 
-/// Ascribes types to the given expression.
-///
-/// In case of an error, the expression is still returned, but the type is set to [`ERROR`].
-fn ascribe_types(
+fn ascribe_literal(force_string: bool, lit: &Literal) -> TypedExpr {
```
syldium, line 175: The `force_string` parameter seems to be too specific. As it is the typing phase, no optimization should be done here.

compiler/src/bytecode.rs
```diff
@@ -1,14 +1,31 @@
+use std::mem::size_of;
+
+/// holds bytecode and contains utility
+/// methods to help compiler write basic instructions
```
syldium, line 4:
```suggestion
/// Holds the currently generated bytecode.
///
/// While the bytecode simply consists of a sequence of bytes, this struct provides
/// methods to emit the correct instructions for multi-byte values.
```

vm/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ if (MSVC)
     set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} /W4 /WX")
 endif()
 
-add_library(vm src/interpreter.cpp src/vm.cpp src/conversions.cpp src/conversions.h)
-add_executable(vm_exe src/main.cpp src/interpreter.cpp src/vm.cpp src/conversions.cpp src/conversions.h)
+add_library(vm src/interpreter.cpp src/vm.cpp src/conversions.cpp src/conversions.h src/memory/operand_stack.cpp src/memory/operand_stack.h)
+add_executable(vm_exe src/main.cpp src/interpreter.cpp src/vm.cpp src/conversions.cpp src/conversions.h src/memory/operand_stack.cpp src/memory/operand_stack.h)
```
syldium, line 15:
```suggestion
add_library(vm src/interpreter.cpp src/vm.cpp src/conversions.cpp src/memory/operand_stack.cpp)
add_executable(vm_exe src/main.cpp)
target_link_libraries(vm_exe vm)
```
Don't list header files here.

vm/src/conversions.h
```diff
@@ -1,12 +1,11 @@
...
 
 #include <cstdint>
-#include <string>
-#include <valarray>
 #include <cstdio>
+#include <cstring>
 #include <memory>
 #include <string>
-#include <cstring>
+#include <valarray>
```
syldium, line 8: Actually used includes:
```h
#include <cstdint>
#include <cstring>
#include <memory>
#include <string>
```

vm/src/interpreter.cpp
```diff
@@ -7,34 +8,47 @@
...
+    OP_FLOAT_TO_STR, // replaces last value of operand stack from float to a string reference
+};
 
 constant_pool::constant_pool(int capacity) {
     strings.reserve(capacity);
     sizes.reserve(capacity);
 }
 
+inline void handle_process_state(int status, const char *bytes, size_t byte_count, unsigned int &ip, OperandStack &stack) {
+    // look ahead for OP_POP_BYTE instruction in order to avoid useless push/pop operations
```
syldium, line 35: The virtual machine should just follow the bytecode and shouldn't make guesses about if the value will be used or not during runtime.

vm/src/interpreter.cpp
```diff
@@ -87,17 +98,18 @@ void run(constant_pool pool, int ip, const char *bytes, size_t size) {
             // Fork and exec the process
             pid_t pid = fork();
             if (pid == 0) {
-                // Execute the process
+                // Execute the process then exit
```
syldium, line 101:
```suggestion
                // Replace the current process with a new process image
```

vm/src/interpreter.cpp
```diff
@@ -87,17 +98,18 @@ void run(constant_pool pool, int ip, const char *bytes, size_t size) {
...
                 execvp(argv[0], argv);
             } else {
                 for (int i = 0; i < frame_size; i++) {
                     delete[] argv[i];
                 }
                 delete[] argv;
+                int status;
```
syldium, line 108:
```suggestion
                int status = 0;
```
Initializing values never hurt :p

vm/src/interpreter.cpp
```diff
@@ -106,51 +118,71 @@ void run(constant_pool pool, int ip, const char *bytes, size_t size) {
...
-            append_str_value(value, stack_local, pool);
             // goto next operation
             ip++;
             break;
         }
+        case OP_IF_NOT_JUMP:
+        case OP_IF_JUMP: {
+            int64_t value = stack.pop_int();
+            size_t then_branch = ntohl(*(size_t *)(bytes + ip + sizeof(char)));
+            // as we are in a shell interpreter thus
```
syldium, line 158: Are we considering that an exit code a boolean as it is? The boolean type also exists, and there must be only one representation.

vm/src/interpreter.cpp
```diff
@@ -106,51 +118,71 @@ void run(constant_pool pool, int ip, const char *bytes, size_t size) {
...
+            // when value is 0 it means that the test or operation succeeded.
+            // the test is reversed if current operation is OP_IF_NOT_JUMP
+            if ((value == 0) == (bytes[ip] == OP_IF_JUMP)) {
```
syldium, line 161: Those checks followed by an XOR work, but overall this single line of code is complicated to read.

vm/src/memory/operand_stack.cpp
```diff
@@ -0,0 +1,60 @@
+//
+// Created by maxime on 6/14/23.
+//
+
```
syldium, line 4:
```suggestion
```

vm/src/memory/operand_stack.cpp
```diff
@@ -0,0 +1,60 @@
...
+    if (current_pos + sizeof(double) >= capacity) {
+        throw OperandStackError("exceeded operand stack capacity");
+    }
+    *(double *)(bytes + current_pos) = d;
+    current_pos += sizeof(double);
+}
+
+void OperandStack::push_string_constant_ref(int64_t s) {
+    if (current_pos + sizeof(int64_t) >= capacity) {
+        throw OperandStackError("exceeded operand stack capacity");
```
syldium, line 30: Write a helper method for those checks, they look all the same.

vm/src/memory/operand_stack.cpp
```diff
@@ -0,0 +1,60 @@
+//
+// Created by maxime on 6/14/23.
+//
+
+#include "operand_stack.h"
+
+OperandStack::OperandStack(size_t len)
+    : bytes{new char[len]},
```
syldium, line 8: Use a smart pointer to allocate the array here.

vm/src/memory/operand_stack.h
```diff
@@ -0,0 +1,45 @@
...
+    double pop_double();
+
+    int64_t pop_string_constant_ref();
+
+    void pop_bytes(size_t size);
+
+    ~OperandStack();
+};
+
+struct OperandStackError : public std::exception {
```
syldium, line 34:
```suggestion
struct OperandStackError : public std::out_of_range {
```

vm/src/vm.cpp
```diff
@@ -26,8 +26,14 @@ constant_pool load_constant_pool(const char *bytes, int *ip) {
...
-    run(std::move(pool), ip, bytes, byte_count);
+extern "C" void exec(const char *bytes, size_t byte_count) {
+    unsigned int constant_pool_bytes = 0;
+    constant_pool pool = load_constant_pool(bytes, &constant_pool_bytes);
+
+    try {
+        run(std::move(pool), bytes + constant_pool_bytes, byte_count - constant_pool_bytes);
+    } catch (std::exception &e) {
+        std::cerr << e.what() << std::endl;
+        exit(1);
```
syldium, line 37: As this is a public API, this shouldn't exit.

syldium (changes requested):

vm/src/interpreter.cpp
Override-6: The bytecode says to pop the value emitted by the `spawn` opcode, the VM can simply avoid pushing the result as the bytecode specifies that it'll be not used

vm/src/interpreter.cpp
Override-6: useless

syldium: What is your guarantee that the function below will actually mutate this value? Having dangling values is a very important thing to avoid in memory unsafe languages like C or C++.

vm/src/memory/operand_stack.h
Override-6: the `OperandStackError` is a kind of error for anything relied to the operand stack; its kind is not bound to any oor stuff

vm/src/interpreter.cpp
Override-6: added a comment line to translate it

vm/src/interpreter.cpp
Override-6: mmh i see indeed, my bad !

vm/src/interpreter.cpp
```diff
@@ -87,17 +97,19 @@ void run(constant_pool pool, int ip, const char *bytes, size_t size) {
...
                 delete[] argv;
+                int status = 0;
                 // Wait for the process to finish
-                waitpid(pid, nullptr, 0);
-                // Pop the stack
-                stack_frame -= frame_size;
+                waitpid(pid, &status, 0);
+
+                // add status to the stack
+                stack.push_int(status);
```
syldium, line 112:
```suggestion
                // Add the exit status to the stack
                stack.push_int(WEXITSTATUS(status));
```
Inserting the status will always result in a truthy value.

vm/src/memory/operand_stack.h
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    template <typename T>
+    T pop() {
+        if (current_pos < sizeof(T)) {
+            throw OperandStackError("operand stack is empty");
+        }
+        current_pos -= sizeof(T);
+        return *(T *)(bytes.get() + current_pos);
+    }
```
syldium, line 65: As this is a private method that is used only in a single source file, place the declaration in the `operand_stack.cpp` file, and just leave the method signatures here.

compiler/src/emit/jump.rs
```diff
@@ -0,0 +1,96 @@
...
+        // then we jump out of the otherwise instruction
+        emitter.emit_code(Opcode::Jump);
+        jump_oo_otherwise_placeholder = Some(emitter.create_placeholder(size_of::<usize>()));
+    }
+
+    emitter.fill_in_ip(jump_to_then_placeholder, emitter.len());
+    emit(&conditional.then, emitter, cp, state);
+    if let Some(jump_placeholder) = jump_oo_otherwise_placeholder {
+        emitter.fill_in_ip(jump_placeholder, emitter.len());
+    }
```
syldium, line 40: There are too many internals used here. Why creating a placeholder here? Why having two `if let Some(_) = otherwise`? This could be rewritten as such:
```rs
    // Evaluate the condition.
    emit(&conditional.condition, emitter, cp, state);
    if conditional.condition.ty == INT {
        emitter.emit_code(Opcode::ConvertIntToByte);
        emitter.emit_byte(1);
        emitter.emit_code(Opcode::BXor);
    }

    // If the condition is false, go to ELSE.
    let jump_to_else = emitter.emit_jump(Opcode::IfNotJump);
    // Evaluate the if branch.
    emit(&conditional.then, emitter, cp, state);
    // Go to END.
    let jump_to_end = emitter.emit_jump(Opcode::Jump);

    // ELSE:
    emitter.patch_jump(jump_to_else);
    if let Some(otherwise) = &conditional.otherwise {
        emit(otherwise, emitter, cp, state);
    }

    // END:
    emitter.patch_jump(jump_to_end);
```

Expose higher level methods in the `Bytecode` struct:
```rs
    /// Emits a jump instruction.
    ///
    /// It returns the index of the offset which is to be patched.
    #[must_use = "the jump address must be patched later"]
    pub fn emit_jump(&mut self, opcode: Opcode) -> usize {
        self.emit_code(opcode);
        self.create_placeholder(size_of::<usize>())
    }

    /// Takes the index of the jump offset to be patched as input, and patches
    /// it to point to the current instruction.
    pub fn patch_jump(&mut self, offset_idx: usize) {
        let ip = self.len();
        self.fill_in_ip(offset_idx, ip);
    }
```

compiler/src/emit/invoke.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    emitter.emit_code(Opcode::Spawn);
+    emitter.bytes.push(arguments.len() as u8);
+
+    if !use_return {
+        // The Spawn operation will push the process's exitcode onto the stack
+        // in order to maintain the stack's size, we instantly pop
+        // the stack if the value isn't used later in the code
+        emitter.emit_code(Opcode::PopByte)
+    }
```
syldium, line 36: This is not specific to process spawns. Consider the following code:
```kt
while true { 1 }
```
The literal value must also be removed from the stack.

compiler/src/emit.rs
```diff
@@ -0,0 +1,101 @@
...
+    cp: &mut ConstantPool,
+    state: &mut EmissionState,
+) {
+    let value = declaration
+        .value
+        .as_ref()
+        .expect("var declaration without initializer not supported");
+    emit(value, emitter, cp, state);
+    emitter.emit_code(Opcode::SetLocal);
+    emitter.bytes.push(declaration.identifier.0 as u8);
```
syldium, line 67:
```suggestion
    if let Some(value) = &declaration.value {
        emit(value, emitter, cp, state);
        emitter.emit_code(Opcode::SetLocal);
        emitter.bytes.push(declaration.identifier.0 as u8);
    }
```
This is a regression.
https://github.com/moshell-lang/moshell/blob/8823ae3e8236d1777fde0e2d7d2a5e48021e00da/compiler/src/lib.rs#L15
On the master branch, the initializer is pattern matched so that it supports variables without initializers. This is only a concern for the analyzer.

analyzer/src/steps/typing.rs
```diff
@@ -122,6 +151,439 @@ fn apply_types_to_source(
...
+            condition: condition.map(Box::new),
+            body: Box::new(body),
+        }),
+        segment: loo.segment(),
+        ty: NOTHING,
+    }
+}
+
+fn ascribe_continue_or_break(
+    e: &Expr,
```
syldium, line 560: Avoid single letter variable names.

syldium (changes requested):

vm/src/memory/operand_stack.h
syldium: This exception is only used when bound checking fails. If more error cases are added in the future, more exceptions types would just be created.

compiler/src/emit/jump.rs
```diff
@@ -0,0 +1,94 @@
...
+pub fn emit_continue(emitter: &mut Bytecode, state: &mut EmissionState) {
+    emitter.emit_code(Opcode::Jump);
+    emitter.emit_instruction_pointer(state.enclosing_loop_start);
+}
+
+pub fn emit_break(emitter: &mut Bytecode, state: &mut EmissionState) {
+    emitter.emit_code(Opcode::Jump);
+    state
+        .enclosing_loop_end_placeholders
+        .push(emitter.create_placeholder(size_of::<usize>()));
```
syldium, line 93:
```suggestion
    state
        .enclosing_loop_end_placeholders
        .push(emitter.emit_jump(Opcode::Jump));
```

compiler/src/emit/jump.rs
```diff
@@ -0,0 +1,94 @@
...
+        emit(&lp.body, emitter, cp, &mut loop_state);
+        emitter.emit_code(Opcode::Jump);
+        emitter.emit_instruction_pointer(start_ip)
+    }
+
+    // fill break placeholders
+    let current_ip = emitter.len();
+    for placeholder in loop_state.enclosing_loop_end_placeholders {
+        emitter.fill_in_ip(placeholder, current_ip)
+    }
```
syldium, line 81:
```suggestion
    // START:
    let loop_start = emitter.current_ip();
    let mut loop_state = EmissionState::in_loop(loop_start);

    if let Some(condition) = &lp.condition {
        // Evaluate the condition.
        emit(condition, emitter, cp, state);
        if condition.ty == INT {
            emitter.emit_code(Opcode::ConvertIntToByte);
            emitter.emit_byte(1);
            emitter.emit_code(Opcode::BXor);
        }
        // If the condition is false, go to END.
        let jump_to_end = emitter.emit_jump(Opcode::IfNotJump);
        loop_state.enclosing_loop_end_placeholders.push(jump_to_end);
    }

    loop_state.enclosing_loop_start = loop_start;

    // Evaluate the loop body.
    emit(&lp.body, emitter, cp, &mut loop_state);
    // Go to START.
    emitter.emit_loop(loop_start);
    // END:
    for jump_to_end in &loop_state.enclosing_loop_end_placeholders {
        emitter.patch_jump(*jump_to_end);
    }
```
Use the newly added methods, and regroup all jump patches in a single place.

compiler/src/emit.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use crate::bytecode::{Bytecode, Opcode};
+use crate::constant_pool::ConstantPool;
+use crate::emit::invoke::emit_process_call;
+use crate::emit::jump::{emit_break, emit_conditional, emit_continue, emit_loop};
+
+mod invoke;
+mod jump;
+
+pub struct EmissionState {
```
syldium, line 14:
```suggestion
#[derive(Debug, Clone, Default)]
pub struct EmissionState {
```
Derive the `Default` trait instead of implementing a custom `new` constructor.

compiler/src/emit.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub enclosing_loop_end_placeholders: Vec<usize>,
+}
+
+impl EmissionState {
+    pub fn new() -> Self {
+        Self {
+            enclosing_loop_start: 0,
+            enclosing_loop_end_placeholders: Vec::new(),
+        }
+    }
```
syldium, line 30:
```suggestion
    /// Create a new emission state.
    pub fn new() -> Self {
        Self::default()
    }

    /// Create a new emission state for a loop.
    pub fn in_loop(loop_start: usize) -> Self {
        Self {
            enclosing_loop_start: loop_start,
            enclosing_loop_end_placeholders: Vec::new(),
        }
    }
```

analyzer/src/steps/typing.rs
```diff
@@ -122,6 +151,439 @@ fn apply_types_to_source(
...
+                relations,
+                diagnostics,
+                env,
+                &w.condition,
+                state.with_local_type(),
+            )),
+            &w.body,
+        ),
+        Expr::Loop(l) => (None, &l.body),
+        _ => unreachable!(),
```
syldium, line 538:
```suggestion
        _ => unreachable!("Expression is not a loop"),
```

analyzer/src/steps/typing.rs
```diff
@@ -74,6 +87,22 @@ impl TypingState {
...
+        }
+    }
+
+    /// Returns a new state with `in_loop` set to false
+    fn without_in_loop(self) -> Self {
+        Self {
+            in_loop: false,
+            ..self
+        }
+    }
```
syldium, line 105:
```suggestion
```

syldium (changes requested):

syldium (approved):
*If* it gets approved, *jump* on the master branch.

PR #110, opened 2023-06-16
Implement type descriptions

syldium:
Types have some sort of capacities, they have methods, can be called or be converted to other types.

While the analyzer have an interesting conceptual model where everything is a method, this becomes a chicken-and-egg problem to implement in a VM. This PR handles that by a listing of native methods to leave the implementation to the host language. The intermediate representation tries to contain every information that the compiler need, and now contains the function and method declaration ID instead of the function name. `Declaration` is a new type, that is either the classical `SourceId`, or the new `NativeId` type.

To facilitate the work of the compiler, the typing phase now also emits every cast/unification it encounters. A cast just change the declared type and does not do any coercion. On the other hand, conversions are done with specific methods, like `to_bool()`, `to_float()` or `to_string()`. Sometimes those methods can implicitly be called in string templates or in process calls. A few types can always implicitly convert to another, it is the case of `ExitCode` to `Bool`, and of `Int` to `Float`.

Everything is a method. This means that binary expressions are just regular method calls, that are now properly checked.

A lot of the diff comes from the added tests and the diagnostic generation. Some work have been done to move them out and reduce the length of the `ascribe_type` function.

Two methods may have the same name on the same type. This PR adds a basic method resolution by always looking for an exact match (no implicit cast). Methods are stored in the `TypedEngine`, as they share a lot with `TypedExpr`s.

Variables can now be assigned. To identify non re-assignable types, the `TypeContext` now also remembers if the variable is immutable or not.


analyzer/src/steps/typing.rs
```diff
@@ -146,67 +155,160 @@ fn ascribe_types(
                 segment: lit.segment.clone(),
             }
         }
+        Expr::TemplateString(tpl) => {
+            if tpl.parts.is_empty() {
+                return TypedExpr {
+                    kind: ExprKind::Literal(LiteralValue::String("".to_owned())),
```
Override-6, line 161:
```suggestion
                    kind: ExprKind::Literal(LiteralValue::String(String::new())),
```

analyzer/src/types/builtin.rs
```diff
@@ -0,0 +1,67 @@
...
+            name_operator_method(*op),
+            MethodType::native(vec![INT], INT, gen.next()),
+        );
+        engine.add_method(
+            FLOAT,
+            name_operator_method(*op),
+            MethodType::native(vec![FLOAT], FLOAT, gen.next()),
+        );
+    }
+    for stringify in [INT, FLOAT] {
```
Override-6, line 49: any primitive should be convertible to string

Override-6 (changes requested):

analyzer/src/types/ty.rs
```diff
@@ -28,14 +35,44 @@ pub enum Type {
...
 
-    /// A callable type, that have a separate environment.
-    Function(SourceId),
+    /// A callable type, that have a separate definition.
+    Function(Definition),
+}
+
+/// A type identifier in a [`TypedEngine`].
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Definition {
```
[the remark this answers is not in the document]
syldium: I was thinking about that, and found that the `Engine` may actually also need this type. The resolution step could theoretically resolve a symbol like `read` to a native function.

analyzer/src/steps/typing.rs
```diff
@@ -146,67 +155,160 @@ fn ascribe_types(
...
+            let rhs = match unify_and_map(
+                rhs,
+                var_ty,
+                &mut exploration.typing,
+                &exploration.engine,
+                state,
+                diagnostics,
+            ) {
+                Ok(rhs) => {
+                    if !var_obj.can_reassign {
```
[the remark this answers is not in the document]
syldium: Functions and some other types are not reassignables, so two diagnostics would be raised instead of one (one because a function cannot be reassigned, and the other because a function does not accept the right-hand side expression).

analyzer/src/steps/typing.rs
```diff
@@ -146,67 +155,160 @@ fn ascribe_types(
...
+                    diagnostics,
+                    env,
+                    part,
+                    state.without_local_type(),
+                );
+                convert_into_string(acc, exploration, diagnostics, state)
+            });
+            let acc = it.next().unwrap();
+            it.fold(acc, |acc, current| {
+                let segment = acc.segment.start..current.segment.end;
```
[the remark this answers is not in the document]
syldium: The first idea is not the best, I was using the accumulated string as a way to express the `add` method. `current.segment` is fine alone.

analyzer/src/steps/typing/lower.rs
```diff
@@ -0,0 +1,84 @@
...
+        return expr;
+    }
+
+    let method_name = match get_converter(into) {
+        Some(method_name) => method_name,
+        None => {
+            panic!(
+                "No converted defined for type `{}`",
+                typing.get_type(into).unwrap()
+            );
```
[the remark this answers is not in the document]
syldium: I could, but technically this function should never be called if `Typing::unify` failed.

analyzer/src/types/ty.rs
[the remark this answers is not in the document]
Override-6: perhaps you may want to move it

analyzer/src/steps/typing.rs
```diff
@@ -166,6 +164,163 @@ fn ascribe_literal(lit: &Literal) -> TypedExpr {
...
+                .unwrap()
+                .type_id,
+        )
+        .unwrap();
+    if actual_type.is_callable() {
+        diagnostics.push(
+            Diagnostic::new(
+                DiagnosticID::TypeMismatch,
+                state.source,
+                format!("Cannot treat function `{}` as a variable", assign.name),
```
Override-6, line 249: this is false, you can access functions as variables, but you can't _assign_ them
```suggestion
                format!("Cannot treat functions as variables for assignement", assign.name),
```
or
```suggestion
                format!("functions cannot be assigned", assign.name),
```

analyzer/src/types/engine.rs
```diff
@@ -108,6 +108,10 @@ impl TypedEngine {
         }
     }
 
+    /// Returns the chunk with the given source id.
+    ///
+    /// If the chunk is not a user defined chunk, [`None`] is returned.
+    /// Use [`Self::get`] to get both user defined and native chunks.
     pub fn get_user(&self, id: SourceId) -> Option<&Chunk> {
         self.entries.get(id.0).and_then(|x| x.as_ref())
```
Override-6, line 116:
```suggestion
        self.entries.get(id.0).as_ref()
```

analyzer/src/types/engine.rs
```diff
@@ -108,6 +108,10 @@ impl TypedEngine {
         }
     }
 
+    /// Returns the chunk with the given source id.
+    ///
+    /// If the chunk is not a user defined chunk, [`None`] is returned.
+    /// Use [`Self::get`] to get both user defined and native chunks.
     pub fn get_user(&self, id: SourceId) -> Option<&Chunk> {
```
Override-6, line 115:
```suggestion
    pub fn get_user_defined_chunk(&self, id: SourceId) -> Option<&Chunk> {
```

compiler/src/emit/native.rs
```diff
@@ -0,0 +1,50 @@
+use crate::bytecode::{Bytecode, Opcode};
+use crate::constant_pool::ConstantPool;
+use crate::emit::{emit, EmissionState};
+use analyzer::relations::NativeId;
+use analyzer::types::hir::TypedExpr;
+
+/// Emits a native sequence of instructions.
+pub(crate) fn emit_native(
```
Override-6, line 8:
```suggestion
pub(crate) fn emit_intrinsics(
```

Override-6 (changes requested):

compiler/src/emit/native.rs
Override-6: for me natives are more VM's implementations, and intrinsics are compiler-defined functions

syldium: An intrinsic function can be implemented directly by the compiler if possible, instead of calling an actual function in a library. An intrinsic function could have two implementations, one that is more performant but enabled only if optimizations are required. See this [definition](https://en.wikipedia.org/wiki/Intrinsic_function):
> Compilers that implement intrinsic functions generally enable them only when a program requests [optimization](https://en.wikipedia.org/wiki/Optimizing_compiler), otherwise falling back to a default implementation provided by the language [runtime system](https://en.wikipedia.org/wiki/Runtime_system) (environment).

Here are emitted native operations, that can only exist in the host VM. They expose C functions that could not be reached if they didn't exist. Adding two ints cannot be done without reaching the host language, here's why this function should be called *native*.

analyzer/src/types/engine.rs
syldium: `Self::get()` already returns a `Chunk`, so this would not be in phase with this other method.

analyzer/src/steps/typing.rs
syldium: Functions are not variables, even outside assignments. They can be assigned to a variable, but are not themselves variables.

compiler/src/emit/native.rs
Override-6: When emitting for example an addition between two ints, The compiler will replace the call by compiler-defined code, which is simply one opcode, the compiler is only bound to its instruction set, and the fact that the vm _interprets_ IADD opcode does not makes `IADD` a kind of native function, moreover the compiler will not fallback to functions implemented in the _native_ language for those binary operators because there are instructions for them which makes the 'fallback' case unreachable.
binary operators are bytecode-level instructions, not native code, and the compiler will detect those methods _intrinsics_ methods to inline them with their bytecode equivalent

syldium: The compiler cannot replace the call, since it cannot be user-defined. Intrinsic methods are methods with fallbacks, that the compiler could treat normally if it wanted to skip them. Here, because of the high level representation, the analyzer technically forgot that is a binary expression that cannot be defined as a function. The trick is to mark those functions as native to signal that the compiler must handle them and cannot emit a regular code. In fact, the whole pipeline is named as *natives* and not *intrinsics*.
Those are terms for the compiler, it doesn't matter if at the end the compiler have a dedicated opcode or just call a named regular function that the VM would need to intercept.

analyzer/src/steps/typing.rs
syldium: Renamed to convey that named types are different from variables.

compiler/src/emit/native.rs
syldium: Renamed to *primitives operations*.

Override-6 (approved):

PR #111, opened 2023-06-20
Enhance cli and repl, migrate to ariadne error reporter, enhance diagnostics, add outputs visualizers

Override-6:
This Pull request focuses on repl/cli inputs and outputs

repl enhancements :
* history support
* carret support in prompt
* indent support in prompt

cli / repl enhancements : 
* lexed tokens visualisation
* parsed AST visualisation
* compiler bytecode visualisation
* clearer, and more precise controls over reports thanks to ariadne migration
* observations through other files and environments now supported
* add `-w` option to set working directory

diagnostics :
* support observation to foreign environments
* possibility to add a tag to observations
* notes supports in diagnostics

sorry for the catastrophic history, i tried to maintain this branch among other branches but this ended up being a total mess, at least i know now 😬


Override-6:
_this pull request will be squashed_

cli/src/main.rs
```diff
@@ -1,31 +1,84 @@
...
-        let content = std::fs::read_to_string(&source)?;
-        let name = source.to_string_lossy().deref().to_string();
-        let source = Source::new(&content, &name);
-        exit(handle_source(source) as i32)
+        exit(run(source, working_dir, config) as i32)
+    }
+    repl(config);
+}
+
+fn update_working_dir(working_dir: Option<PathBuf>) -> PathBuf {
```
syldium, line 35: An erroneous working directory should return an error and not `exit(1)` the program.

analyzer/src/diagnostic.rs
```diff
@@ -82,26 +82,49 @@ pub enum DiagnosticID {
...
+        self.tag = Some(tag);
+        self
+    }
+}
+
+#[derive(Clone, PartialEq, Debug)]
+pub enum ObservationTag {
+    InFault,
+    Expected,
+    Declaration,
```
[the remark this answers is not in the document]
Override-6: `Expected` is for segments from which we expected something, `Declaration` is for segments with no particular expectation but got observed because they are relevant in the diagnostic (ex: `parameter is declared here`)

cli/src/main.rs
Override-6: it's just an utility function, it is declared in `main` and is not meant to be used in any other way that its currently used

syldium: So they are three main issues:
- Just panic if the working directory cannot be changed. If you want to customize it more, it is a sign that you shouldn't use blindly `exit(1)`.
- Having the working directory existing doesn't mean that `set_current_dir()` will success if the directory has restrictive permissions.
- If the user wants to change the directory, they can just change it before invoking the cli. It is a source path, not a working directory path.

analyzer/src/types/engine.rs
```diff
@@ -58,6 +59,9 @@ pub struct Chunk {
 
     /// The return type of the chunk.
     pub return_type: TypeId,
+
+    /// Segment of declaration
+    pub declaration_segment: Option<SourceSegment>,
```
syldium, line 64: Isn't the chunk the whole segment? We might actually need to differentiate string sources from environment sources, as a single string can create multiple environments.

cli/src/cli.rs
```diff
@@ -1,99 +1,91 @@
...
+
+    /// Activate compiler output visualisation
+    #[arg(short = 'B', long)]
+    pub(crate) bytecode: bool,
 }
 
-#[derive(Default)]
-struct RawImporter<'a> {
-    imported_modules: HashMap<Name, Expr<'a>>,
+pub struct Configuration {
```
syldium, line 34: An option to avoid executing the code would be useful.

cli/src/display_bytecode.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    let byte_padding = bytes_len.to_string().len();
+
+    while ip < bytes_len {
+        let opcode: Opcode = bytes[ip].try_into().expect("not an opcode");
+        macro_rules! get_usize {
+            () => {
+                usize::from_be_bytes((&bytes[ip..ip + 8]).try_into().unwrap())
+            };
+        }
```
syldium, line 24:
```suggestion
        macro_rules! get_fixed_slice {
            ($size:expr) => {
                bytes[ip..ip + $size].try_into().unwrap()
            };
        }
        macro_rules! get_usize {
            () => {
                usize::from_be_bytes(get_fixed_slice!(8))
            };
        }
```
You can use `get_fixed_slice!` to get float values.

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+use context::source::Source;
+
+use crate::fix_ariadne_report;
+
+pub fn render_diagnostic<'a>(
+    diagnostic: analyzer::diagnostic::Diagnostic,
+    source_supplier: impl Fn(SourceId) -> Source<'a>,
+) -> io::Result<String> {
+    let mut colors = ColorGenerator::new();
+    let mut colormap = HashMap::new();
```
syldium, line 21: colormap is never inserted, and is only used for `ObservationTag::Other` that shouldn't have a magic color value.

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+        )
+    } else {
+        (
+            format!("warn[W{:04}]", diagnostic.identifier.code()),
+            Color::Yellow,
+        )
+    };
+
+    let labels = diagnostic.observations.into_iter().map(|o| {
+        let source = o.foreign_env.unwrap_or(diagnostic.source);
```
syldium, line 56: It seems like only observations have a source. Maybe remove the `SourceId` from `Diagnostic` and let `Observation` have a `pub source: SourceId` field.

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+    if let Some(note) = note {
+        builder = builder.with_note(colorize_message(note, Color::Cyan))
+    }
+
+    let cache = sources
+        .into_values()
+        .map(|s| (s.name, ariadne::Source::from(s.source)))
+        .collect::<HashMap<_, _>>();
+
+    let cache = CacheMap { cache };
```
syldium, line 97: Reuse the same cache for all diagnotics.

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    let mut buf = Vec::new();
+    builder.finish().write_for_stdout(cache, &mut buf)?;
```
syldium, line 100:
```suggestion
    builder.finish().write(cache, &mut buf)?;
```
The diagnotic will be written to `stderr`

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+use yansi::Paint;
+
+use context::source::Source;
+
+use crate::fix_ariadne_report;
+
+pub fn render_diagnostic<'a>(
+    diagnostic: analyzer::diagnostic::Diagnostic,
+    source_supplier: impl Fn(SourceId) -> Source<'a>,
+) -> io::Result<String> {
```
syldium, line 19: Don't create an intermediate string and take an `impl std::io::Write`.

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+            Some(head.clone())
+        } else {
+            let notes_agg = tail
+                .iter()
+                .fold(format!("\n- {head}"), |acc, str| format!("{acc}\n- {str}"));
+            Some(notes_agg)
+        }
+    }
+    agg
+}
```
syldium, line 119:
```suggestion
fn agg_strings(mut strings: Vec<String>) -> Option<String> {
    if let Some(last) = strings.pop() {
        if strings.is_empty() {
            Some(last)
        } else {
            let mut builder = String::new();
            for note in strings {
                write!(builder, "- {note}\n").unwrap();
            }
            write!(builder, "- {last}").unwrap();
            Some(builder)
        }
    } else {
        None
    }
}
```

cli/src/formatted_parse_error.rs
```diff
@@ -0,0 +1,41 @@
...
+                .with_message("Start")
+                .with_color(Color::Green)
+                .with_order(1),
+        ),
+        _ => builder,
+    };
+
+    let mut buf = Vec::new();
+
+    builder.finish().write_for_stdout(
```
syldium, line 34:
```suggestion
    builder.finish().write(
```

cli/src/formatted_parse_error.rs
```diff
@@ -0,0 +1,41 @@
+use std::io;
+
+use ariadne::{Color, Config, Label, Report, ReportKind};
+
+use crate::fix_ariadne_report;
+use context::source::Source;
+use parser::err::{ParseError, ParseErrorKind};
+
+pub fn render_parse_error(source: Source, error: ParseError) -> io::Result<String> {
```
syldium, line 9: Take a `impl std::io::Write`

cli/src/main.rs
```diff
@@ -1,31 +1,84 @@
...
+        );
+        exit(1);
+    }
+
+    std::env::set_current_dir(working_dir.clone()).expect("could not set current dir");
+    working_dir
+}
+
+/// https://github.com/zesterer/ariadne/issues/51
+pub fn fix_ariadne_report(report_str: String) -> String {
```
syldium, line 59: Given that you resolved this issue, this workaround is no longer necessary, and we can wait for a proper release of `ariadne`.

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
+use std::collections::HashMap;
+use std::process::exit;
+
+type REPLEditor = Editor<(), DefaultHistory>;
+
+#[derive(Default)]
+struct REPLImporter<'a> {
+    #[allow(clippy::vec_box)]
+    // Box is used to ensure that the reference behind is still valid after vector's realloc
+    sources: Vec<Box<OwnedSource>>,
```
syldium, line 34: This is not necessary to box the source, as only the string pointers will be exposed and not the `Source` itself.

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
+    let chars = line_clone.chars();
+    for c in chars {
+        if !c.is_whitespace() {
+            return indent;
+        }
+        line.remove(0);
+        indent.push(c);
     }
+    indent
 }
```
syldium, line 110:
```suggestion
```
Stripping whitespaces is the job of the parser.

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
+    let mut indent_prefix = String::new();
+    loop {
+        let line = editor.readline_with_initial(&prompt_prefix, (&indent_prefix, ""));
+        let mut line = match line {
+            Ok(line) => line,
+            Err(ReadlineError::Interrupted) | Err(ReadlineError::Eof) => exit(1),
+            e => e.expect("error when reading next line from editor"),
+        };
+
+        indent_prefix = strip_indent(&mut line);
```
syldium, line 126:
```suggestion
```
This doesn't make any sense, how the prefix is determined from leading spaces?

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
             continue; // Silently ignore incomplete input
         }
 
-        return Some(OwnedSource::new(
-            source.source.to_string(),
-            source.name.to_string(),
-        ));
+        editor
+            .add_history_entry(source.source.to_string())
+            .expect("terminal has no history");
```
syldium, line 145:
```suggestion
            .ok();
```
Don't panic just because the line cannot be appended to the history.

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
+        return;
+    }
+
+    if config.parser_visualization {
+        display_exprs(&report.expr)
+    }
+
+    let expr = Expr::Block(Block {
+        expressions: report.expr,
+        segment: 0..0,
```
syldium, line 196:
```suggestion
        segment: source.segment(),
```

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
+    let source = importer.take_source(source);
+    let name = Name::new(source.name);
+
+    if config.lexer_visualisation {
+        display_tokens(lex(source.source))
+    }
+
+    let report = parse(source);
+
+    let source = source;
```
syldium, line 178:
```suggestion
```

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
+    }
+
+    let mut bytecode = Vec::new();
+    let root_expr = typed_engine
+        .get(Definition::User(SourceId(0)))
+        .map(|c| match c {
+            CodeEntry::User(c) => &c.expression,
+            _ => unreachable!(),
+        })
+        .unwrap();
```
syldium, line 222:
```suggestion
    let root_expr = &typed_engine.get_user(SourceId(0)).unwrap().expression;
```

cli/src/runner.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    let mut bytecode = Vec::new();
+    let root_expr = typed_engine
+        .get(Definition::User(SourceId(0)))
+        .map(|c| match c {
+            CodeEntry::User(c) => &c.expression,
+            _ => unreachable!(),
+        })
+        .unwrap();
```
syldium, line 61:
```suggestion
    let root_expr = &typed_engine.get_user(SourceId(0)).unwrap().expression;
```
This is also duplicated with the REPL runner.

cli/src/runner.rs
```diff
@@ -0,0 +1,179 @@
...
+use parser::err::ParseReport;
+use parser::parse;
+
+use crate::cli::{display_exprs, display_tokens, execute, Configuration};
+use crate::display_bytecode::display_bytecode;
+use crate::formatted_diagnostic::render_diagnostic;
+use crate::formatted_parse_error::render_parse_error;
+use crate::source_importer::FileSourceImporter;
+
+pub(crate) fn run(source: PathBuf, working_dir: PathBuf, config: Configuration) -> bool {
```
syldium, line 26: The returned bool is inverted, did you mean to return true in case of success? If it is for the `exit(code as i32)`, it should be inverted here.

parser/src/err.rs
```diff
@@ -107,17 +107,14 @@ pub struct ParseReport<'a> {
...
 
-    /// Indicates that the end of the input has been reached too early and
-    /// that the parser is waiting for more input.
-    ///
-    /// This flag is set only when a block delimiter is opened, but not closed,
-    /// and there was no actual error in the input.
-    pub stack_ended: bool,
+    /// contains the delimiter stack that were not closed,
+    /// This ParseReport is considered as valid if the delimiter stack is empty
+    pub delimiter_stack: Vec<TokenType>,
```
syldium, line 112: Expose the last token in the stack, but not the entire stack.

syldium (changes requested):

cli/src/cli.rs
Override-6: why not indeed

cli/src/display_bytecode.rs
Override-6: not worth it as it is used only one time for PushFloat

cli/src/formatted_diagnostic.rs
Override-6: yes that was the goal of ObservationTag::Other, group observations by "other" tags and then the cli would have a different color for those groups

cli/src/formatted_diagnostic.rs
Override-6: We would lose the origin of the diagnostic

syldium: The diagnostic doesn't have the source, it is the definition, that is an `Observation`.

cli/src/repl.rs
Override-6: if the vector is reallocated, the OwnedSource will be moved, `Source` points to the OwnedSource, thus if moved the references will be invalid

cli/src/formatted_diagnostic.rs
Override-6: We still lose the origin of the diagnostic

cli/src/repl.rs
Override-6: This is meant to keep the same indent for multi line prompts and keep an history without the indent

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    let mut color_msg = String::new();
+
+    for cap in regex.captures_iter(&msg) {
+        if let Some(mach) = cap.get(1) {
+            let start = mach.start();
+            let end = mach.end();
+
+            color_msg = color_msg.add(&msg[last..start]);
+            let colorized = Paint::new(&msg[start..end]).fg(color).to_string();
```
syldium, line 135: The color is always applied even if the terminal doesn't support colors.

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+            let notes_agg = tail
+                .iter()
+                .fold(format!("\n- {head}"), |acc, str| format!("{acc}\n- {str}"));
+            Some(notes_agg)
+        }
+    }
+    agg
+}
+
+/// parses given string and colorizes segments between anti quotes (`) with given color
```
syldium, line 121:
```suggestion
/// Parses a given string and colorizes segments between backticks (`) with a given color.
```

cli/src/formatted_diagnostic.rs
```diff
@@ -0,0 +1,156 @@
...
+    color_msg.add(&msg[last..])
+}
+
+struct CacheMap<'a> {
+    cache: HashMap<&'a str, ariadne::Source>,
+}
+
+impl<'a> Cache<&'a str> for CacheMap<'a> {
+    fn fetch(&mut self, id: &&'a str) -> Result<&ariadne::Source, Box<dyn Debug + '_>> {
+        Ok(self.cache.get(id).unwrap())
```
syldium, line 150:
```suggestion
        Ok(self.cache.get(id).expect("Cache not correctly populated"))
```

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
+pub fn repl(config: Configuration) {
+    let mut editor: REPLEditor =
+        DefaultEditor::new().expect("unable to instantiate terminal editor");
+    editor.set_color_mode(ColorMode::Enabled);
+    editor.set_history_ignore_dups(true).unwrap();
+    editor.set_history_ignore_space(true);
+    editor.bind_sequence(
+        Event::KeySeq(vec![KeyEvent(KeyCode::Char('u'), Modifiers::ALT)]),
+        EventHandler::from(Cmd::Undo(1)),
+    );
```
syldium, line 77:
```suggestion
```

cli/src/repl.rs
```diff
@@ -1,47 +1,230 @@
...
 
 /// Indefinitely prompts a new expression to the stdin,
 /// displaying back the errors if any and the formed AST
-pub fn prompt() {
-    while let Some(source) = parse_input() {
-        handle_source(source.as_source());
+pub fn repl(config: Configuration) {
+    let mut editor: REPLEditor =
+        DefaultEditor::new().expect("unable to instantiate terminal editor");
+    editor.set_color_mode(ColorMode::Enabled);
```
syldium, line 71:
```suggestion
```
Don't enforce colors.

syldium (changes requested):

analyzer/src/diagnostic.rs
```diff
@@ -82,33 +82,54 @@ pub enum DiagnosticID {
...
 pub struct Observation {
-    /// Observed segment
+    /// Observed source
+    pub source: SourceId,
+    /// Observed source's segment
     pub segment: SourceSegment,
     /// An optional help string to complete the observation
     pub help: Option<String>,
+    /// An optional tag to group observations
+    pub tag: Option<ObservationTag>,
```
syldium, line 92: The tag could just be `InFault` by default, and not being `None`.

cli/src/cli.rs
```diff
@@ -1,99 +1,91 @@
...
+    for token in tokens {
+        print!(
+            "{:10}: '{}', ",
+            format!("{:?}", token.token_type).bright_blue(),
+            token.value.white()
+        );
+        count += 1;
+        if count % 5 == 0 {
+            println!()
         }
```
syldium, line 70:
```suggestion
```

cli/src/cli.rs
```diff
@@ -1,99 +1,91 @@
...
+    println!()
+}
 
-    let mut stdout = stderr();
-    let had_errors = !diagnostics.is_empty();
-    for diagnostic in diagnostics {
-        display_diagnostic(source, diagnostic, &mut stdout)
-            .expect("IO errors when reporting diagnostic")
+pub(crate) fn display_exprs(exprs: &Vec<Expr>) {
+    println!("{}", "AST: ".cyan());
```
syldium, line 76: This doesn't check if colors are supported.

cli/src/repl.rs
```diff
@@ -1,47 +1,210 @@
...
+use crate::render::diagnostic::render_diagnostic;
+use crate::render::parse_error::render_parse_error;
+use crate::render::SourcesCache;
+
+type REPLEditor = Editor<(), DefaultHistory>;
+
+#[derive(Default)]
+struct REPLImporter<'a> {
+    #[allow(clippy::vec_box)]
+    // Box is used to ensure that the reference behind is still valid after vector's realloc
```
syldium, line 35:
```suggestion
```

cli/src/repl.rs
```diff
@@ -1,47 +1,210 @@
...
-    for line in lines {
-        let line = line.expect("couldn't read line from stdin");
+    let mut prompt_prefix = "=> ".to_string();
+    let mut indent_prefix = "";
+    loop {
+        let line = editor.readline_with_initial(&prompt_prefix, (indent_prefix, ""));
+        let mut line = match line {
+            Ok(line) => line,
+            Err(ReadlineError::Eof) => exit(0),
+            Err(ReadlineError::Interrupted) => exit(1),
```
syldium, line 106: Receiving an interruption signal is not an error.

cli/Cargo.toml
```diff
@@ -6,16 +6,21 @@ edition = "2021"
...
 parser = { path = "../parser" }
 analyzer = { path = "../analyzer" }
 ast = { path = "../ast" }
+lexer = { path = "../lexer" }
 compiler = { path = "../compiler" }
 
-miette = { version = "5.9.0", features = ["fancy"] }
+owo-colors = "3.0.0"
+concolor = "0.1.1"
+rustyline = { version = "11.0.0", features = ["custom-bindings"] }
```
syldium, line 18:
```suggestion
rustyline = "11.0.0"
```

cli/src/runner.rs
```diff
@@ -0,0 +1,176 @@
...
+
+use crate::cli::{display_exprs, display_tokens, execute, Configuration};
+use crate::disassemble::display_bytecode;
+use crate::render::diagnostic::render_diagnostic;
+use crate::render::parse_error::render_parse_error;
+use crate::render::SourcesCache;
+use crate::source_importer::FileSourceImporter;
+
+/// returns true if any error occurred
+pub(crate) fn run(source: PathBuf, working_dir: PathBuf, config: Configuration) -> bool {
```
syldium, line 28:
```suggestion
pub(crate) fn run(source: PathBuf, source_dir: PathBuf, config: Configuration) -> bool {
```
The `bool` is still inverted.

syldium (changes requested):

cli/src/cli.rs
```diff
@@ -1,99 +1,90 @@
...
-        let types = apply_types(&result.engine, &result.relations, &mut diagnostics);
-        if diagnostics.is_empty() {
-            execute(types);
-            return false;
+pub(crate) fn display_tokens(tokens: Vec<Token>) {
+    println!("Lexer tokens:");
+    println!("\t- {} tokens lexed from input", tokens.len());
+
+    let mut count = 0;
+    for token in tokens {
```
syldium, line 60:
```suggestion
    for (count, token) in tokens.iter().enumerate() {
```

cli/src/cli.rs
```diff
@@ -1,99 +1,90 @@
...
-
-    let result = resolve_all(name, &mut importer);
-
-    let mut diagnostics = result.diagnostics;
-    if diagnostics.is_empty() {
-        let types = apply_types(&result.engine, &result.relations, &mut diagnostics);
-        if diagnostics.is_empty() {
-            execute(types);
-            return false;
+pub(crate) fn display_tokens(tokens: Vec<Token>) {
```
syldium, line 55:
```suggestion
pub(crate) fn display_tokens(tokens: &[Token]) {
```

cli/src/cli.rs
```diff
@@ -1,99 +1,90 @@
...
     }
+    println!()
+}
 
-    let mut stdout = stderr();
-    let had_errors = !diagnostics.is_empty();
-    for diagnostic in diagnostics {
-        display_diagnostic(source, diagnostic, &mut stdout)
-            .expect("IO errors when reporting diagnostic")
+pub(crate) fn display_exprs(exprs: &Vec<Expr>) {
```
syldium, line 74:
```suggestion
pub(crate) fn display_exprs(exprs: &[Expr]) {
```

cli/src/disassemble.rs
```diff
@@ -0,0 +1,151 @@
+use compiler::bytecode::Opcode;
+
+pub(crate) fn display_bytecode(bytes: &[u8]) {
+    println!("Bytecode:");
+    let (ip, constants) = display_constant_pool(bytes);
+    println!();
+    display_byte_instructions(constants, &bytes[ip..])
+}
+
+fn display_byte_instructions(constants: Vec<String>, bytes: &[u8]) {
```
syldium, line 10:
```suggestion
fn display_byte_instructions(constants: &[String], bytes: &[u8]) {
```

cli/src/disassemble.rs
```diff
@@ -0,0 +1,151 @@
...
+            Opcode::GetLocal => {
+                print!("<local @{}>", bytes[ip]);
+                ip += 1;
+            }
+            Opcode::SetLocal => {
+                print!("<constant @{}>", bytes[ip]);
+                ip += 1;
+            }
+            Opcode::Spawn => {
+                print!("<arguments stack size {}>", bytes[ip]);
```
syldium, line 69:
```suggestion
                print!("<arity {}>", bytes[ip]);
```

cli/src/main.rs
```diff
@@ -1,31 +1,28 @@
...
-    .expect("miette options setup");
+    let config = Configuration::from(cli.clone());
 
     if let Some(source) = cli.source {
-        let content = std::fs::read_to_string(&source)?;
-        let name = source.to_string_lossy().deref().to_string();
-        let source = Source::new(&content, &name);
-        exit(handle_source(source) as i32)
+        let working_dir = std::env::current_dir().expect("No working dir set");
+        exit(!run(source, working_dir, config) as i32)
```
syldium, line 25:
```suggestion
        let source_dir = std::env::current_dir().expect("No working dir set");
        exit(!run(source, source_dir, config) as i32)
```

cli/src/render.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use ariadne::Cache;
+
+use analyzer::relations::SourceId;
+use context::source::Source;
+
+pub(crate) mod diagnostic;
+pub(crate) mod parse_error;
+
+pub struct SourcesCache<S> {
```
syldium, line 12: Please document what the supplier does, and/or a type constraint to it.

parser/src/err.rs
```diff
@@ -107,21 +107,18 @@ pub struct ParseReport<'a> {
     /// This may be empty if the parsing process succeeded.
     pub errors: Vec<ParseError>,
 
-    /// Indicates that the end of the input has been reached too early and
-    /// that the parser is waiting for more input.
-    ///
-    /// This flag is set only when a block delimiter is opened, but not closed,
-    /// and there was no actual error in the input.
-    pub stack_ended: bool,
+    /// contains the last delimiter that were not closed,
```
syldium, line 110:
```suggestion
    /// Holds the last opening delimiter that was not closed.
    ///
```

parser/src/err.rs
```diff
@@ -107,21 +107,18 @@ pub struct ParseReport<'a> {
...
+    /// This ParseReport is considered as valid if the unclosed delimiter is none
```
syldium, line 111:
```suggestion
    /// If a value is present, the input has been reached too early and
    /// that the parser is waiting for more input.
```

syldium (changes requested):
Merge the `render/diagnostic.rs` and `render/parse_error.rs` files together.

cli/src/render.rs
```diff
@@ -0,0 +1,200 @@
...
+use ariadne::{Cache, Label, Report, ReportKind};
+use concolor::Stream;
+use yansi::{Color, Paint};
+
+use analyzer::relations::{ObjectId, SourceId};
+use context::source::Source;
+use parser::err::{ParseError, ParseErrorKind};
+use std::fmt::Write as FmtWrite;
+
+/// contains the sources bound to each [SourceId]
```
syldium, line 16:
```suggestion
/// Associates a [`SourceId`] to the appropriate [`ariadne::Source`] type.
///
/// This cache is used to integrate with the ariadne crate, and is not used for
/// any other purpose.
```

cli/src/render.rs
```diff
@@ -0,0 +1,200 @@
...
+            None => {
+                let source = (self.supplier)(*id);
+                self.names.insert(*id, source.name.to_string());
+                let src = self
+                    .sources
+                    .entry(source.name.to_string())
+                    .or_insert_with(|| ariadne::Source::from(source.source));
+                Ok(src)
+            }
+        }
```
syldium, line 57: Map directly the `SourceId` type to the `ariadne::Source` type:
```rs
pub struct SourcesCache<'a, S>
where
    S: Fn(SourceId) -> Source<'a>,
{
    supplier: S,
    names: HashMap<SourceId, String>,
    sources: HashMap<SourceId, ariadne::Source>,
}
```
```suggestion
        let source = (self.supplier)(*id);
        self.names.insert(*id, source.name.to_string());
        Ok(self
            .sources
            .entry(*id)
            .or_insert_with(|| ariadne::Source::from(source.source)))
```

cli/src/render.rs
```diff
@@ -0,0 +1,200 @@
...
+            Color::Yellow,
+        )
+    };
+
+    let labels = diagnostic.observations.into_iter().map(|o| {
+        let mut label = Label::new((o.source, o.segment));
+        cache.fetch(&o.source).expect("Could not fetch source");
+
+        if o.tag == ObservationTag::Expected {
+            label = label.with_order(100)
```
syldium, line 118:
```suggestion
            label = label.with_order(100);
```

cli/src/render.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        let color = get_color(o.tag);
+        label = label.with_color(color);
+
+        if let Some(help) = o.help {
+            label = label.with_message(colorize_message(help, color))
```
syldium, line 125:
```suggestion
            label = label.with_message(colorize_message(help, color));
```

cli/src/repl.rs
```diff
@@ -1,47 +1,205 @@
...
+
+impl<'a, const N: usize> From<[(Name, Expr<'a>); N]> for REPLImporter<'a> {
+    fn from(value: [(Name, Expr<'a>); N]) -> Self {
+        Self {
+            imported_modules: HashMap::from(value),
+            sources: Vec::new(),
+        }
+    }
+}
+
```
syldium, line 44:
```suggestion
```

cli/src/repl.rs
```diff
@@ -1,47 +1,205 @@
...
+impl<'a> REPLImporter<'a> {
+    pub fn take_source(&mut self, source: OwnedSource) -> Source<'a> {
+        self.sources.push(source);
+        let src = self.sources[self.sources.len() - 1].as_source();
+        unsafe {
+            // SAFETY: The sources will never be removed from the self.sources vector as the REPLImporter's
+            // role is to be the owner of the user's sources.
+            // The reference behind Box does not change and is valid for the lifetime of the importer.
```
syldium, line 58:
```suggestion
            // SAFETY: A source is owned by the importer and are never removed.
            // A Source is the reference version to the Strings inside the OwnedSource,
            // so if the OwnedSource moves, the strings are still valid.
            // 'a is used here to disambiguate the lifetime of the source and the mutable borrow.
```

cli/src/runner.rs
```diff
@@ -0,0 +1,174 @@
...
+
+impl<'a> RunnerImporter<'a> {
+    fn new(root: PathBuf) -> Self {
+        Self {
+            source_importer: FileSourceImporter::new(root),
+            errors: Vec::new(),
+        }
+    }
+
+    fn import_expr(&mut self, source: Source<'a>) -> Option<Expr<'a>> {
```
syldium, line 162:
```suggestion
    fn parse_expr(&mut self, source: Source<'a>) -> Option<Expr<'a>> {
```

cli/src/source_importer.rs
```diff
@@ -0,0 +1,88 @@
...
+            return Ok(unsafe {
+                // SAFETY: A source is owned by the importer.
+                // 'a is used here to disambiguate the lifetime of the source and the mutable borrow.
+                std::mem::transmute::<Source, Source<'a>>(source.as_source())
+            });
+        }
+        let source = read_to_string(&path)
+            .or_else(|_| {
+                path.pop();
+                path.set_extension(MOSHELL_EXTENSION);
```
syldium, line 76: This behavior has huge flaws. The analyzer will already try to import the parent name if a source cannot be found. read_to_string returns an Result<String, io::Error>, that can designate a file that does not exist or a file that cannot be read, for instance. Those two cases should be treated differentially.

syldium (changes requested):

analyzer/src/steps/typing/function.rs
```diff
@@ -391,26 +393,30 @@ pub(super) fn type_method<'a>(
 fn diagnose_arg_mismatch(
     typing: &Typing,
     source: SourceId,
+    fn_def: Definition,
     param: &Parameter,
```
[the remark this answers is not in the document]
Override-6: this'll do the trick for now, its not the purpose of this pr to change `Parameter` it implies too much diff to change the structure

cli/src/render.rs
Override-6: No because this would make multiple source id bound to the same source, so duplicate the sources

analyzer/src/types/ty.rs
```diff
@@ -64,6 +64,7 @@ pub struct FunctionType {
 /// A function parameter.
 #[derive(Clone, Debug, PartialEq)]
 pub struct Parameter {
+    pub(crate) fn_def: Definition,
     pub(crate) segment: Option<SourceSegment>,
```
syldium, line 68: If a `segment` is used, then the `fn_def` is always `Definition::User`.

Define the `Parameter` struct as:
```rs
/// A location in an engine.
#[derive(Clone, Debug, PartialEq)]
pub struct Location {
    pub(crate) definition: SourceId,
    pub(crate) segment: SourceSegment,
}

/// A function parameter.
#[derive(Clone, Debug, PartialEq)]
pub struct Parameter {
    pub(crate) location: Option<Location>,
    pub(crate) ty: TypeId,
}
```

You can then use it at your advantage with the help of pattern matching:
```rs
    if let Some(Location {
        definition: source,
        segment,
    }) = &param.location
    {
        diagnostic = diagnostic.with_observation(
            Observation::new(segment.clone(), *source)
                .with_help("Parameter is declared here")
                .with_tag(ObservationTag::Declaration),
        )
    }
```

cli/src/cli.rs
```diff
@@ -1,99 +1,90 @@
...
-
-    let result = resolve_all(name, &mut importer);
-
-    let mut diagnostics = result.diagnostics;
-    if diagnostics.is_empty() {
-        let types = apply_types(&result.engine, &result.relations, &mut diagnostics);
-        if diagnostics.is_empty() {
-            execute(types);
-            return false;
+pub(crate) fn display_tokens(tokens: Vec<Token>) {
```
syldium, line 57:
```suggestion
pub(crate) fn display_tokens(tokens: &[Token]) {
```

cli/src/repl.rs
```diff
@@ -1,47 +1,208 @@
...
+fn display_diagnostics(diagnostics: Vec<Diagnostic>, source: Source) {
+    let mut cache = SourcesCache::new(|_| source);
+    for diagnostic in diagnostics {
+        render_diagnostic(diagnostic, &mut cache, &mut stderr())
+            .expect("IO errors when reporting diagnostic");
+    }
+}
+
+/// Parses and display errors / diagnostics coming from the given source.
+/// Returning true if the source had at least one error or diagnostic.
```
syldium, line 144: This function does not return a boolean.

syldium (changes requested):

PR #112, opened 2023-06-20
Add mutative operations to compiler

Override-6:


analyzer/src/steps/typing.rs
```diff
@@ -711,9 +711,12 @@ fn ascribe_call(
             convert_into_string(expr, exploration, diagnostics, state)
         })
         .collect::<Vec<_>>();
+
+    let ty = if state.local_type { EXIT_CODE } else { NOTHING };
```
syldium, line 715: It is not up to the analyzer to check if an expression is used. Every unused return must be tracked later, as it is a concern for a lot of expressions: `while true { 1 }`, `while true { 1 + 1 }`, `while true { "here" }`, val o = 5; `while true { $o }`...

`local_type` is used only to determine if the nested expression must coerce to a same type.

compiler/src/emit.rs
```diff
@@ -85,6 +84,20 @@ fn emit_block(
...
+    emitter: &mut Bytecode,
+    cp: &mut ConstantPool,
+    state: &mut EmissionState,
+) {
+    emit(&assignment.rhs, emitter, cp, state);
+    if let Symbol::Local(id) = assignment.identifier {
+        emitter.emit_set_local(id.0 as u8);
+        return;
+    }
+    unimplemented!("cannot support external variables assignations")
```
syldium, line 98:
```suggestion
    match assign.identifier {
        Symbol::Local(id) => emitter.emit_set_local(id.0 as u8),
        Symbol::External(_) => {
            unimplemented!("External variable assignations are not implemented yet")
        }
    }
```
Be exhaustive in match patterns.

syldium (changes requested):

analyzer/src/steps/typing.rs
Override-6: We does not have this phase yet, and `is_local_type` is clearly used for expressions whose return type is retained. for now the compiler can simply check if an expression that usually returns something is typed as UNIT to pop the value, because if useless values arent popped, the operand stack gets overflowed

Override-6: will finally move this detection in the compiler in a next pull request

syldium (approved):

PR #113, opened 2023-06-21
Pop unused expression values

Override-6:
This pull request, as discussed in #112, adds a new state during compilation phase (`EmissionState::use_values`) which helps the compiler tracking unused values pushed from expressions.
Those check are needed to keep an operand stack as small as possible and with a deterministic size


syldium (approved):
Please reformat before merging.

Issue #115, opened 2023-06-22
Do not report typecheck diagnostics relied with any `Error` type


Override-6:
Do not push diagnostics that involves the `Error` special type in order to report only top level issues
here's an example of the issue
![image](https://github.com/moshell-lang/moshell/assets/67809972/837ad8b5-43bc-4e41-9869-b28d67cfb8ba)
The second error could be avoided.

Another example : 
![image](https://github.com/moshell-lang/moshell/assets/67809972/d17a309f-dc90-4e63-9e7d-eb2b4885fbd3)

PR #117, opened 2023-06-23
Add comparison instructions to the VM

syldium:
Supporting arithmetic and comparisons operations provides a useful base to create interesting loops. Comparisons emits a one-byte sized type, which mean that it cannot be used in a variable, as a variable always holds eight bytes for now.

The modulo operator has been removed for the `Float` type, as it is only applicable to integer division.

This PR also brings some fixes for #115, by allowing the `Error` type to unify with any type.


vm/src/interpreter.cpp
```diff
@@ -121,7 +216,20 @@ void run(constant_pool pool, const char *bytes, size_t size) {
             pid_t pid = fork();
             if (pid == 0) {
                 // Replace the current process with a new process image
-                execvp(argv[0], argv);
+                if (execvp(argv[0], argv) == -1) {
+                    for (int i = 0; i < frame_size; i++) {
+                        delete[] argv[i];
+                    }
+                    delete[] argv;
```
Override-6, line 223: no need to free if you exit, and this fragment is duplicated three times

Override-6 (changes requested):

vm/src/interpreter.cpp
syldium: Abusing the fact that exiting a process automatically is not worth it. Regarding the repetition for each error case, it is relatively small, but a small refactoring could effectively help here.

Override-6 (approved):

PR #120, opened 2023-06-29
Add function invocations to VM

Override-6:
This pull requests adds a call stack notion to the virtual machine, and adds support for function invocation and declaration in virtual machine and compiler.


vm/src/memory/call_stack.cpp
```diff
@@ -0,0 +1,119 @@
+#include "call_stack.h"
+
+/// Contains all the runtime information about a frame
```
syldium, line 3:
```suggestion
/**
 * Contains all the runtime information about a frame.
 */
```

vm/src/memory/locals.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <algorithm>
+#include <cstdint>
+#include <errors.h>
+#include <string>
+
+/// The size of a local
```
syldium, line 8: Incorrect description.

vm/src/vm.h
```diff
@@ -9,7 +9,7 @@
 /**
  * Executes the given Moshell bytecode.
  *
- * The bytecode will be immediately executed without any sanity checks.
+ * The bytecode will be immediately executed.
```
syldium, line 12:
```suggestion
 * The bytecode will be immediately executed without any sanity checks.
```
This is still true, if the generated bytecode is invalid, this is still undefined behavior.

vm/src/memory/constant_pool.cpp
```diff
@@ -15,29 +14,24 @@ const std::string *read_string(ByteReader &reader, strings_t &strings) {
 }
 
 ConstantPool load_constant_pool(ByteReader &reader, strings_t &strings) {
-    try {
-        // Read the number of strings on a single byte
-        uint32_t count = ntohl(reader.read<uint32_t>());
+    // Read the number of strings on a single byte
```
syldium, line 17:
```suggestion
    // Read the number of strings on four bytes
```

vm/src/vm.cpp
```diff
@@ -1,18 +1,23 @@
 #include "byte_reader.h"
+#include "definitions/module_definition.h"
 #include "interpreter.h"
-#include "memory/constant_pool.h"
-#include "memory/strings.h"
+#include <cxxabi.h>
```
syldium, line 4:
```suggestion
```
Don't depend on libstdc++ internals.

vm/src/vm.cpp
```diff
@@ -1,18 +1,23 @@
...
+
+        run_module(module_def, strings);
 
-        run(pool, bytes + reader.position(), byte_count - reader.position(), strings);
     } catch (const std::exception &e) {
-        std::cerr << e.what() << '\n';
+        int status;
+        char *exception_type = abi::__cxa_demangle(typeid(e).name(), nullptr, 0, &status);
+        std::cerr << exception_type << ": " << e.what() << std::endl;
+        free(exception_type);
```
syldium, line 21:
```suggestion
        std::cerr << e.what() << '\n';
```
Don't depend on libstdc++ internals.

syldium (changes requested):

compiler/src/bytecode.rs
```diff
@@ -21,120 +28,213 @@ impl Bytecode {
...
+        self.bytes.extend(value.to_be_bytes());
+    }
+
+    /// emits a constant pool reference, which is an unsigned 32 bits integer
+    pub fn emit_constant_ref(&mut self, constant: u32) {
+        self.emit_u32(constant);
+    }
+
+    /// Fills an instruction pointer at given instruction pointer in the byte array
+    pub fn patch_u32_placeholder(&mut self, placeholder: Placeholder, value: u32) {
```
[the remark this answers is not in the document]
Override-6: placeholders can be of any length, even if the `Bytecode` impl only provide u32 placeholders i wanted to explicit the type of placeholder

vm/src/vm.h
syldium: A sanity check is not a runtime check. The fact that a jump address is not verified, that a local size is not verified and a lot of things implies that there are no sanity checks.

compiler/src/constant_pool.rs
```diff
@@ -1,14 +1,15 @@
 use indexmap::IndexSet;
 
-/// Contains the constants defined in constant pool
+/// Contains the constants defined in a module constant pool
```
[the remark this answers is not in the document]
Override-6: one module = one bytecode file

compiler/src/bytecode.rs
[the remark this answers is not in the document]
syldium: This is still false to call it a `u32_placeholder` as it is a four-byte placeholder.

compiler/src/lib.rs
```diff
@@ -1,35 +1,179 @@
...
+    let return_bytes_count: u8 = get_type_stack_size(chunk.return_type).into();
+    bytecode.emit_byte(return_bytes_count);
+
+    // if the function returns void, we won't the compiler to emit useless values
+    let use_value = return_bytes_count != 0;
+
+    // emit instruction count placeholder
+    let instruction_count = bytecode.emit_u32_placeholder();
+
+    let start = bytecode.len();
```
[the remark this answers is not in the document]
Override-6: no the Bytecode struct is just a structure to emit basic types

compiler/src/constant_pool.rs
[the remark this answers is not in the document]
syldium: A module is not a unit of work. Modules are semantic separations, while files are technical separations.

analyzer/src/steps/typing/lower.rs
```diff
@@ -49,7 +49,9 @@ pub(super) fn call_convert_on(
     state: TypingState,
 ) -> TypedExpr {
     // If the expression is already of the needed type, we don't need to do anything.
-    if expr.ty.is_err() || into.is_err() || expr.ty == into {
+    // if one of the two expressions is NOTHING, we do perform any conversion following rule `A U Nothing => A`
+    if expr.ty.is_err() || into.is_err() || expr.ty == into || expr.ty == NOTHING || into == NOTHING
+    {
```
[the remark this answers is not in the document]
Override-6: indeed, you also cant convert Nothing to Anything, this diff is wrong from the beginning

compiler/src/bytecode.rs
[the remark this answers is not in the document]
Override-6: it's not just a "four byte" placeholder, it's a `u32` placeholder. For now we only have one type of placeholded value, but i still want to underline the fact that the placeholder is an `u32`.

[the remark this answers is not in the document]
syldium: The bytecode holds bytes, nothing more. The `u32` type is a representation for something that can be inserted in the byte vector, but does not convey any meaning. The type is also present in the method signature, this kinda repeat the same information. I would consider that placeholders are something that are only meaningful in the struct itself, and are re-exposed with more context than the internal type in new public methods. This is an address placeholder, so I would call it `patch_adress()`.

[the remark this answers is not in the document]
Override-6: its a _u32_ placeholder, we want to encapsulate the emission and patch of placeholders and not just consider them as "bytes holes" because it is the meaning of the `Bytecode` structure. `emit_u32_placeholder` comes with `patch_u32_placeholder`. I want to explicit that we manipulate `u32` placeholders because as you said placeholders are just an area wating for some undetermined bytes.

compiler/src/emit.rs
```diff
@@ -16,30 +22,28 @@ mod native;
...
 
     // if set to false, the compiler will avoid emitting literals, var references or will
     // instantly pop values returned from functions, methods and process calls
-    // we don't use values by default
     pub use_values: bool,
+
+    // if set to true, the compiler will place the value it usually pushes in operand stack in first local's index, regardless
+    // of the LocalsLayout.
+    // This is how the function returns their values according to bytecode specifications
+    pub is_returning_value: bool,
```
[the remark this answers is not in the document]
Override-6: indeed its much simpler

compiler/Cargo.toml
```diff
@@ -8,4 +8,6 @@ edition = "2021"
 [dependencies]
 analyzer = { path = "../analyzer" }
 ast = { path = "../ast" }
-indexmap = "1.9.3"
+indexmap = "2.0.0"
+num-traits = "0.2"
+
```
syldium, line 13:
```suggestion
```

compiler/src/bytecode.rs
```diff
@@ -21,120 +32,203 @@ impl Bytecode {
...
-        pos
+    /// emits an instruction pointer
+    fn emit_instruction_pointer(&mut self, ip: u32) {
+        self.bytecode.emit_u32(ip);
     }
 
     /// Emits a jump instruction.
     ///
-    /// It returns the index of the offset which is to be patched.
+    /// It returns the Placeholder of the offset which is to be patched
```
syldium, line 185:
```suggestion
    /// It returns the [`Placeholder`] adress of the offset which is to be patched.
```

compiler/src/emit.rs
```diff
@@ -15,31 +18,25 @@ mod native;
 #[derive(Debug, Clone, Default)]
 pub struct EmissionState {
     // the start instruction position of the enclosing loop
-    // set to 0 if there is no loop
-    pub enclosing_loop_start: usize,
+    // set_bytes to 0 if there is no loop
```
syldium, line 21:
```suggestion
    // set to 0 if there is no loop
```

compiler/src/emit.rs
```diff
@@ -15,31 +18,25 @@ mod native;
...
+    pub enclosing_loop_start: u32,
+
     // All the placeholders waiting for the end of the loop.
     // When the loop compilation ends, all those placeholder are filled with the
     // first instruction pointer after the loop.
-    pub enclosing_loop_end_placeholders: Vec<usize>,
+    pub enclosing_loop_end_placeholders: Vec<Placeholder>,
 
-    // if set to false, the compiler will avoid emitting literals, var references or will
+    // if set_bytes to false, the compiler will avoid emitting literals, var references or will
```
syldium, line 29:
```suggestion
    // if set to false, the compiler will avoid emitting literals, var references or will
```

compiler/src/lib.rs
```diff
@@ -1,35 +1,153 @@
...
+    // patch instruction count placeholder
+    let instruction_byte_count =
+        u32::try_from(instructions.current_ip()).expect("too much instructions");
+    bytecode.patch_u32_placeholder(instruction_count, instruction_byte_count);
+
+    // as the locals also contains parameters and the return value, the
+    // length is the maximum of those three values
+    let locals_length = locals
+        .byte_count()
+        .max(return_bytes_count as u32)
```
syldium, line 130:
```suggestion
```

parser/src/err.rs
```diff
@@ -110,7 +110,7 @@ pub struct ParseReport<'a> {
     /// Indicates that the end of the input has been reached too early and
     /// that the parser is waiting for more input.
     ///
-    /// This flag is set only when a block delimiter is opened, but not closed,
+    /// This flag is set_bytes only when a block delimiter is opened, but not closed,
```
syldium, line 113:
```suggestion
    /// This flag is set only when a block delimiter is opened, but not closed,
```

parser/src/moves.rs
```diff
@@ -110,8 +110,8 @@ where
     }
 }
 
-///Move to next token if its type is in the given set
-/// * `set` - the set of TokenType to satisfy
+///Move to next token if its type is in the given set_bytes
+/// * `set_bytes` - the set_bytes of TokenType to satisfy
```
syldium, line 114:
```suggestion
///Move to next token if its type is in the given set
/// * `set` - the set_bytes of TokenType to satisfy
```

vm/src/memory/operand_stack.h
```diff
@@ -1,44 +1,96 @@
...
+
+    /**
+     * pops `n` bytes
+     * @throws OperandStackUnderflowError if the operand stack does not have enough bytes to pop
+     */
+    const char *pop_bytes(size_t n);
+
+    /**
+     * advances, without checking for stack overflow, the position of the operand stack.
+     * The call of this method must be JUSTIFIED
```
syldium, line 91:
```suggestion
     * The caller of this method must ensure that the given size will not overflow the stack.
```

vm/src/memory/strings.h
```diff
@@ -4,8 +4,8 @@
 #include <unordered_set>
 
 /**
- * The `strings_t` type is a simple set that contains all the strings during the lifetime
+ * The `strings_t` type is a simple set_bytes that contains all the strings during the lifetime
```
syldium, line 7:
```suggestion
 * The `strings_t` type is a simple set that contains all the strings during the lifetime
```

vm/src/memory/strings.h
```diff
@@ -4,8 +4,8 @@
...
  * of the virtual machine.
- * All string constants and runtime strings are contained in this set.
- * */
+ * All string constants and runtime strings are contained in this set_bytes.
```
syldium, line 9:
```suggestion
 * All string constants and runtime strings are contained in this set.
```

vm/src/vm.h
```diff
@@ -9,7 +13,7 @@
 /**
  * Executes the given Moshell bytecode.
  *
- * The bytecode will be immediately executed without any sanity checks.
+ * The bytecode will be immediately executed without any sanity check.
```
syldium, line 16:
```suggestion
 * The bytecode will be immediately executed without any sanity checks.
```

vm/src/vm.h
```diff
@@ -1,6 +1,10 @@
 #pragma once
 #include <cstddef>
 
+#if UINTPTR_MAX > 0xFFFFFFFFFFFFFFFFu // 64 bits
+#error "VM only supports architectures less than 64 bits architectures"
+#endif
+
```
syldium, line 7: Why a larger address space would be an issue?

analyzer/src/steps/typing.rs
```diff
@@ -1320,19 +1325,27 @@ mod tests {
...
+                Diagnostic::new(
+                    DiagnosticID::CannotInfer,
+                    SourceId(1),
+                    "Return type is not inferred for block functions"
+                )
+                .with_observation(Observation::with_help(
+                    find_in(content, "var b: Int = $a"),
+                    "Returning `Unit`"
+                ))
+                .with_help("Try adding an explicit return type to the function")
```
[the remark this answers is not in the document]
Override-6: Unit indeed, i'll make the change

compiler/src/lib.rs
Override-6: the return value is stored in the locals. 
```kt
fun foo() = "hello"
val x = foo()
echo $x
```
This would lead to an error if we remove the line

vm/src/vm.h
Override-6: because the bytecode defines a reference on 8 bytes (64 bits). which supports 64/32/16 etc architectures but cannot support larger addresses. We do not want to make the bytecode platform dependent and the VM has no function signatures to know the size of its stack/parameter/return value

compiler/src/lib.rs
syldium: The issue is that this line is duplicated just below :p

Override-6: indeed what i meant in the code was
```rs
locals
        .byte_count()
        .max(parameters_bytes_count)
        .max(return_bytes_count as u32);
```

compiler/src/bytecode.rs
```diff
@@ -21,120 +32,203 @@ impl Bytecode {
...
+    }
+
     pub fn emit_code(&mut self, code: Opcode) {
-        self.bytes.push(code as u8);
+        self.bytecode.emit_byte(code as u8)
+    }
+
+    pub fn emit_pop(&mut self, size: ValueStackSize) {
+        let pop_opcode = match size {
+            ValueStackSize::Zero => panic!("Cannot pop zero sized types"),
```
syldium, line 101:
```suggestion
            ValueStackSize::Zero => return,
```
Writing `fun a() = {}; a()` should be valid.

compiler/src/lib.rs
```diff
@@ -1,35 +1,153 @@
...
-    for constant in pool.strings {
-        writer.write_all(&constant.len().to_be_bytes())?;
-        writer.write_all(constant.as_bytes())?;
+fn write(
+    writer: &mut impl Write,
+    bytecode: &Bytecode,
+    pool: &ConstantPool,
+) -> Result<(), io::Error> {
+    write_constant_pool(pool, writer)?;
+    writer.write_all(&bytecode.bytes())
```
syldium, line 141:
```suggestion
    writer.write_all(bytecode.bytes())
```

compiler/src/lib.rs
```diff
@@ -1,35 +1,153 @@
...
+
+    let mut instructions = Instructions::wrap(bytecode);
+    let mut locals = LocalsLayout::fixed_count(chunk_vars.all_vars().len());
+
+    for (id, param) in chunk.parameters.iter().enumerate() {
+        locals.initialize_space(LocalId(id), param.ty)
+    }
+
+    let mut state = EmissionState::default();
+    state.use_values = use_value;
```
syldium, line 109:
```suggestion
    let mut state = EmissionState {
        use_values: use_value,
        ..EmissionState::default()
    };
```

syldium (changes requested):

compiler/src/locals.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// reserves the space in the locals depending on the stack size needed by the given type
+    pub fn initialize_space(&mut self, id: LocalId, ty: TypeId) {
+        if self.indexes[id.0].is_some() {
+            return;
+        }
+        let size: u8 = get_type_stack_size(ty).into();
+        self.indexes[id.0] = Some(self.len);
+        self.len += size as u32;
+    }
```
syldium, line 28:
```suggestion
    /// Reserves the space in the locals depending on the stack size needed by the given type.
    ///
    /// Different initialization orders will result in different indexes.
    ///
    /// # Panics
    /// Panics if the local id is out of bounds.
    pub fn initialize_space(&mut self, id: LocalId, stack_size: ValueStackSize) {
        let size: u8 = stack_size.into();
        self.indexes[id.0] = Some(self.total_size);
        self.total_size += size as u32;
    }
```
Do not silently ignore function parameters. Since this struct is responsible for determining the layout, externalize the process of calculating the size of a type.

compiler/src/locals.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// get first index allocated for given local identifier
+    pub fn get_index(&self, id: LocalId) -> u32 {
+        self.indexes[id.0].unwrap_or_else(|| {
+            panic!(
+                "Local {} is unknown or has no initialized space in locals layout",
+                id.0
+            )
+        })
+    }
```
syldium, line 38:
```suggestion
    /// Get starting byte index allocated for the given local.
    ///
    /// Returns [`None`] if the local is size has not yet been initialized.
    ///
    /// # Panics
    /// Panics if the local id is out of bounds.
    pub fn get_index(&self, id: LocalId) -> Option<u32> {
        self.indexes[id.0]
    }
```

vm/src/definitions/function_definition.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <vector>
```
syldium, line 5:
```suggestion
```

vm/src/definitions/module_definition.cpp
```diff
@@ -0,0 +1,67 @@
+#include "byte_reader.h"
+#include "bytecode_unit.h"
+#include "conversions.h"
+#include "definitions/function_definition.h"
+#include "memory/constant_pool.h"
+#include "memory/locals.h"
+
+std::unordered_map<const std::string *, function_definition>
+load_functions_definitions(ByteReader &reader, const ConstantPool &pool) {
```
syldium, line 9:
```suggestion
load_function_definitions(ByteReader &reader, const ConstantPool &pool) {
```

vm/src/memory/call_stack.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include "definitions/function_definition.h"
+#include "locals.h"
+#include "operand_stack.h"
+
+/// Thrown when the memory allocated for a thread call stack
+/// is exceeded
+class StackOverflowError : public MemoryError {
+public:
+    explicit StackOverflowError(std::string message) : MemoryError{message} {}
```
syldium, line 11:
```suggestion
/**
 * Thrown when the memory allocated for a thread call stack is exceeded.
 */
class StackOverflowError : public MemoryError {
public:
    explicit StackOverflowError(std::string message) : MemoryError{std::move(message)} {}
```
Don't mix Rust documentation style with Doxygen documentation style.

vm/src/memory/locals.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+void Locals::set_ref(uint64_t r, size_t at) {
+    set<uint64_t>(r, at);
+}
+void Locals::set_bytes(const char *data, size_t size, size_t at) {
+    check_capacity(at, size, "updating");
+    memcpy((void *)bytes, data, size);
+}
+
+inline void Locals::check_capacity(size_t at, size_t space_size, std::string action) const {
```
syldium, line 31:
```suggestion
inline void Locals::check_capacity(size_t at, size_t space_size, std::string_view action) const {
```
Don't allocate a debug string each time.

vm/src/memory/locals.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+/**
+ * Encapsulates the allocated locals area of a stack frame
+ */
+class Locals {
+    /**
+     * encapsulated bytes in the stack frame
+     */
+    const char *bytes;
```
syldium, line 23:
```suggestion
    char *const bytes;
```
The `const` should apply to the pointer, not the `char`.

vm/src/memory/locals.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+void Locals::set_byte(char b, size_t at) {
+    set<char>(b, at);
+}
+void Locals::set_ref(uint64_t r, size_t at) {
+    set<uint64_t>(r, at);
+}
+void Locals::set_bytes(const char *data, size_t size, size_t at) {
+    check_capacity(at, size, "updating");
+    memcpy((void *)bytes, data, size);
```
syldium, line 28:
```suggestion
    memcpy(static_cast<void *>(bytes), data, size);
```

vm/src/interpreter.cpp
```diff
@@ -159,279 +221,332 @@ bool apply_comparison(Opcode code, double a, double b) {
...
+/**
+ * Will run a frame until it returns or pushes a new method inside the call_stack
+ * @return true if this function returned because the current frame has ended, or false if it returned because it pushed a new frame
+ */
+bool run_frame(runtime_state &state, stack_frame &frame, CallStack &call_stack, const char *instructions, size_t instruction_count) {
+    const ConstantPool &pool = state.pool;
+
+    // the instruction pointer
+    size_t &ip = *frame.instruction_pointer;
+    OperandStack &operands = frame.operands;
```
syldium, line 247: I have small doubts about calling it *operands* instead of *stack*. While it adds unnecessary diff, it also loses the fact that this is a stack.

vm/src/interpreter.cpp
```diff
@@ -159,279 +221,332 @@ bool apply_comparison(Opcode code, double a, double b) {
...
+            // pop the current frame.
+            call_stack.pop_frame();
+
+            if (call_stack.is_empty()) {
+                // the root method has returned
+                break;
+            }
+            stack_frame caller_frame = call_stack.peek_frame();
+
+            // JUSTIFY: the returned value is placed at the very first index of locals.
```
syldium, line 521:
```suggestion
            // SAFETY: the returned value is placed at the very first index of locals.
```

vm/src/interpreter.cpp
```diff
@@ -159,279 +221,332 @@ bool apply_comparison(Opcode code, double a, double b) {
...
+            // locals of the callee frame and operands of the caller frame are adjacent in memory thanks to
+            // call stack's layout. This operation will transfer the return value in the operand stack of the caller
+            // by advancing the caller's operands of the declared byte count in the callee's function definition.
+            //
+            // JUSTIFY: This also cannot cause UB as a minimal frame size is equal to `sizeof(frame_headers)`,
+            // that contains ints and pointers, which is greater than the maximum return byte count allowed of `sizeof(uint64_t)`.
+            // stack overflow is impossible here because the callee frame would already have caused it earlier.
```
syldium, line 528:
```suggestion
            // The returned value is already guaranteed to not overflow the stack, as it is already there.
```

analyzer/src/environment/variables.rs
```diff
@@ -77,10 +77,16 @@ impl Variables {
...
+    #[cfg(debug_assertions)]
     pub fn all_vars(&self) -> &[Variable] {
         &self.locals.vars
     }
 
+    /// returns the
+    pub fn var_count(&self) -> usize {
+        self.locals.vars.len()
+    }
+
```
syldium, line 89:
```suggestion
```

analyzer/src/environment/variables.rs
```diff
@@ -77,10 +77,16 @@ impl Variables {
     /// This exposes their current state, which is only interesting for debugging.
     /// Use [`Variables::find_reachable`] to lookup any variable during the collection phase,
     /// or [`Variables::find_exported`] to lookup an exported variable after the collection phase.
+    #[cfg(debug_assertions)]
```
syldium, line 80:
```suggestion
```
This method is not for debug builds only, as it could also be used for visualization purposes.

syldium (changes requested):

vm/src/interpreter.cpp
Override-6: it is the stack for operands.

analyzer/src/environment/variables.rs
Override-6: this is what documentation says

syldium: The documentation states that is for debugging purposes, not for debug builds. Conditional compilation also is something to avoid, as integration tests need to treat the binaries as they are, and not recompile it.

vm/src/interpreter.cpp
syldium: Operand stacks is a term that the JVM uses, while a stack is a more universal term for something to work with. Instructions do *push* and *pop* from the stack, they do not *place* operands like we would *place* locals. On the other hand, this is true that there are more than one type of stack.

analyzer/src/environment/variables.rs
Override-6: if it's only interesting for debugging then it's shall only be used when debugging. 
tests already compiles with debug_assertions set to true thus this not invalidates their build

syldium: That's not the case. Execute `cargo test --release` to test with a release build, and the method will not be present, despite compiling the tests.

cli/src/cli.rs
```diff
@@ -60,16 +61,16 @@ pub fn handle_source(source: Source) -> bool {
         segment: 0..0,
     });
 
-    let name = Name::new("<module>");
+    let name = Name::new("<repl>");
```
syldium, line 64:
```suggestion
    let name = Name::new("<module>");
```
This is also used while reading a file.

compiler/src/lib.rs
```diff
@@ -1,35 +1,154 @@
...
+    let instruction_byte_count = instructions.current_ip();
+    bytecode.patch_u32_placeholder(instruction_count, instruction_byte_count);
+
+    // as the locals also contains parameters and the return value, the
+    // length is the maximum of those three values
+    let locals_length = locals
+        .byte_count()
+        .max(parameters_bytes_count)
+        .max(return_bytes_count as u32);
+    bytecode.patch_u32_placeholder(locals_byte_count, locals_length)
```
syldium, line 133:
```suggestion
    let locals_size = locals.byte_count();
    bytecode.patch_u32_placeholder(locals_byte_count, locals_size);
```
The parameters are already inserted in the layout, and the return value can be directly moved to the right place on the stack instead of depending on the locals.

syldium (approved):

PR #124, opened 2023-07-08
Import multiple files

syldium:
The analyzer has the ability to import multiple sources, but was not implemented for real files.

Importing files
===============

This PR distinguishes three possible outputs if fetching a source from a `Name` fails:

- The source doesn't exist, but a parent may exist. In this case, return `Ok(None)`, to continue trying importing a similar `Name`.
- The source do exist, but is not usable. This might be due to file permissions or to a parsing error. In that case, return `Err(ImportError)` to stop the resolution here.
- The source is intelligible, return `Ok(Some(_))`.

The command line interface is now able to read files and display parse errors that could errors. This can be done either using the `-s <file>` option or in the REPL (still not able to remember previously executed lines).

A file importer looks for in the current working directory for files that ends with the `.msh` extension, but may have exceptions to load direct paths that do not follow this convention.

Mapping diagnostics to the appropriate source code
==================================================

Diagnostics are related to sources, that do not correlate with the ID attributed to the imported files. A single source file may be splitted into multiple `SourceId`s, shifting every index to the right. To get back to the original source code from a `SourceId`, something needed to keep track of the `ContentId` of the source. Since the engine held the origins, it now also remember the original source the expression came from.

Originally, it was up to the `Diagnostic` struct to have the source. Because a single diagnostic can now refer to even different files, it became clear that it is to each `Observation` to store the `SourceId`. While updating all the constructor usages, this PR tries to get on track to differentiate *erroneous* locations from *context* locations in observations.

### DIFF
--- a/analyzer/src/diagnostic.rs
+++ b/analyzer/src/diagnostic.rs
@@ -4,6 +4,7 @@ use context::source::SourceSegment;
 
 use crate::relations::SourceId;
 
+#[non_exhaustive]
 #[derive(PartialEq, Debug, Assoc)]
 #[func(pub fn code(&self) -> u16)]
 #[func(pub fn critical(&self) -> bool { false })]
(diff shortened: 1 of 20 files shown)


Override-6 (changes requested):
The use of `Result<Option<_>, ZeroSizedType>` in `Importer::import` should be replaced by an enum with three variants in order to be more straightforward.

Override-6:
Also i'm not really convinced with the choice of "label" and "underline" words for diagnostics, i think that we should not choose vocabulary related with graphic reports

syldium:
> Also i'm not really convinced with the choice of "label" and "underline" words for diagnostics, i think that we should not choose vocabulary related with graphic reports

Indeed, this is kinda misleading. I went for rewriting the `underline` constructor with an `From` impl, and renamed `label` to `message`. Let me know it is clearer.

Override-6 (approved):

PR #127, opened 2023-07-13
Add native functions invocation support

Override-6:
With the aim to lowerize the bytecode definition level, and thus to only have `ref`, `int`, `byte` and `floats` values, this pull request implements native functions support in order to move all string related opcodes (`OP_INT_TO_STR`, `OP_FLOAT_TO_STR`, `OP_CONCAT`, `OP_STR_EQ`) into native functions.

## Native functions invocations
Native functions needs to have the following signature : `void (OperandStack &caller_stack, runtime_state &state)`. 
They access input data and (possibly) returns computed value through the `caller_stack`. They also have an access to the strings heap, to push in new strings.

definition of `OP_INT_TO_STR`, `OP_FLOAT_TO_STR`, `OP_CONCAT`, `OP_STR_EQ` opcodes were moved in native functions defined in `stdlib_natives.cpp`

### The strings heap
The `strings_t` set has been wrapped into a `StringsHeap` class to encapsulate strings insertion.


vm/src/memory/strings.cpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "memory/strings.h"
+
+const std::string &StringsHeap::insert(std::string str) {
+    auto [it, _] = strings.insert(std::make_unique<std::string>(str));
+    return *it->get();
```
syldium, line 7:
```suggestion
const std::string &StringsHeap::insert(std::string str) {
    auto [it, _] = strings.insert(std::make_unique<std::string>(std::move(str)));
    return **it;
```

syldium (changes requested):

syldium (approved):

PR #128, opened 2023-07-18
Fix symbol capture in collection phase

Override-6:
the commit https://github.com/moshell-lang/moshell/commit/b37a0f5216e1ed9eabb4f809cc904da6e359cf7c did an undesirable regression which made capture resolution through non-direct parents not working.

This pull request fixes this issue by making the collector instantly attach created environments in the engine, and then work with environments inside the engine.


syldium (approved):
I wonder if the `attach` and the `track` methods could be merged. Nice catch!

PR #129, opened 2023-07-21
Handle function captures

Override-6:
This pull request improves the notion of reference on bytecode/VM side and uses this improvement to make the compiler able to handle var captures within functions.

## Bytecode improvements
Current bytecode specification uses `OP_GET_*` and `OP_SET_*` opcodes to get/set a local value. 
With the need of accessing and updating referenced values, the `OP_GET_*` and `OP_SET_*` no longer have parameters and now pops a reference from operand stack to access and/or update a value.
Thus, the `OP_PUSH_LOCAL_REF` has been introduced and pushes on the operand stack a reference to the given local index.
The `OP_PUSH_STRING` has been refactored to `OP_PUSH_CONSTANT_REF` to match the changes.

before: 
```
val a = 5
$a
```
compiled to :
```cpp
OP_PUSH_INT 5
OP_SET_INT 0 //access local at index 0 and set value

OP_GET_INT 0 //access local at index 0 and get value
```
and now compiles to
```cpp
OP_PUSH_INT 5
OP_PUSH_LOCAL_REF 0 //push address of local 0
OP_SET_INT //pop reference, pop value and set value at reference

OP_PUSH_LOCAL_REF 0 //push address of local 0
OP_GET_INT //pop reference, push accessed value from reference.
```

## LocalsLayout
Locals indexes has been split in two parts, a Vector of indexes for locals, and a HashMap for captures, binding their `ResolvedSymbol` with their start position in the locals's area.

## Other changes
- Replaced `externals: HashMap` with `externals: IndexMap` in `Variables` in order to iterate over them in a source-dependant order.
- Replaced use of `Symbol` in HIR with a new `Var` structure that already contains the `ResolvedSymbol` in `Var::Capture` variant, which lighten the use of `relations` by the compiler. 
-

### DIFF
--- a/analyzer/src/environment/variables.rs
+++ b/analyzer/src/environment/variables.rs
@@ -1,5 +1,5 @@
-use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use indexmap::map::Entry;
+use indexmap::IndexMap;
 use std::fmt::{Display, Formatter};
 
 use crate::name::Name;
@@ -31,7 +31,7 @@ pub struct Variables {
 
     /// Relations with external variables.
     /// The key is the variable Names, where value is the relation with another external environment's [Locals]
-    externals: HashMap<Name, RelationId>,
+    externals: IndexMap<Name, RelationId>,
 }
 
 impl Variables {
@@ -79,6 +79,15 @@ impl Variables {
     /// or [`Variables::find_exported`] to lookup an exported variable after the collection phase.
     pub fn all_vars(&self) -> &[Variable] {
         &self.locals.vars
+    }
+
+    /// returns an iterator over all variables, with their local identifier
+    pub fn iter(&self) -> impl Iterator<Item = (LocalId, &Variable)> {
+        self.locals
+            .vars
+            .iter()
+            .enumerate()
+            .map(|(i, v)| (LocalId(i), v))
     }
 
     /// Iterates over all the exported variables, local to the environment.
--- a/compiler/src/emit.rs
+++ b/compiler/src/emit.rs
@@ -1,6 +1,6 @@
 use analyzer::engine::Engine;
-use analyzer::relations::{Definition, Relations, SourceId};
-use analyzer::types::hir::{Declaration, ExprKind, TypeId, TypedExpr, Var};
+use analyzer::relations::{Definition, SourceId};
+use analyzer::types::hir::{Declaration, ExprKind, TypedExpr, Var};
 use ast::value::LiteralValue;
 
 use crate::bytecode::{Instructions, Opcode, Placeholder};
@@ -82,17 +82,11 @@ fn emit_literal(literal: &LiteralValue, instructions: &mut Instructions, cp: &mu
     };
 }
 
-fn emit_ref(var: Var, ref_type: TypeId, instructions: &mut Instructions, locals: &LocalsLayout) {
-    let size = ref_type.into();
-    instructions.emit_get_local(var, size, locals);
-}
-
 #[allow(clippy::too_many_arguments)]
 fn emit_declaration(
     declaration: &Declaration,
     instructions: &mut Instructions,
     engine: &Engine,
-    relations: &Relations,
     cp: &mut ConstantPool,
     locals: &mut LocalsLayout,
     state: &mut EmissionState,
@@ -106,7 +100,6 @@ fn emit_declaration(
             Var::Local(declaration.identifier),
             instructions,
             engine,
-            relations,
             cp,
             locals,
             state,
@@ -120,7 +113,6 @@ fn emit_block(
     exprs: &[TypedExpr],
     instructions: &mut Instructions,
     engine: &Engine,
-    relations: &Relations,
     cp: &mut ConstantPool,
     locals: &mut LocalsLayout,
     state: &mut EmissionState,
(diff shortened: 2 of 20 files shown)


compiler/src/locals.rs
```diff
@@ -1,20 +1,30 @@
...
 
 use crate::r#type::ValueStackSize;
 
 /// contains the different index per local value allocated in the locals area
 pub struct LocalsLayout {
     /// the start indexes of bound Locals
-    indexes: Vec<Option<u32>>,
+    values_indexes: Vec<Option<u32>>,
+    /// the start indexes of bound external values
+    external_refs_indexes: HashMap<ResolvedSymbol, u32>,
```
syldium, line 14: The whole point of captures is to retrieve external symbols, that are firstly references, then copies. You can't keep track of that here, you need to follow along if the referred variable is dropped (because its scope ended) to transform the upvalue into a copy.

compiler/src/locals.rs
```diff
@@ -23,20 +33,36 @@ impl LocalsLayout {
...
+        self.values_indexes[id.0] = Some(self.len);
         self.len += size as u32;
     }
 
+    /// Reserves the space in the eternal's reference
+    ///
+    /// Different initialization orders will result in different indexes.
+    pub fn set_external_ref_space(&mut self, symbol: ResolvedSymbol) {
+        match self.external_refs_indexes.entry(symbol) {
+            Entry::Occupied(_) => {}
```
syldium, line 47: Don't silence errors.

compiler/src/lib.rs
```diff
@@ -71,22 +77,111 @@ pub fn compile(
...
+            .filter(|(_, v)| v.ty == TypeInfo::Function)
+        {
+            let func_fqn = &env.fqn.appended(Name::new(&func.name));
+            let (func_id, _) = engine.find_environment_by_name(func_fqn).unwrap();
+
+            resolve(func_id, engine, relations, captures, externals, chunks)
+        }
+
+        chunks.insert(chunk_id);
+        externals.retain(|symbol| !chunks.contains(&symbol.source));
```
syldium, line 141: Those lines don't seem to make sense. Why would a symbol not be external if already seen? If not, shouldn't the filtering be done only one time at the end?

compiler/src/emit.rs
```diff
@@ -68,151 +82,135 @@ fn emit_literal(literal: &LiteralValue, instructions: &mut Instructions, cp: &mu
...
 #[allow(clippy::too_many_arguments)]
 fn emit_declaration(
     declaration: &Declaration,
     instructions: &mut Instructions,
-    typing: &Typing,
     engine: &Engine,
     cp: &mut ConstantPool,
     locals: &mut LocalsLayout,
     state: &mut EmissionState,
+    captures: &mut Captures,
```
syldium, line 93:
```suggestion
    captures: &Captures,
```
In your current implementation, you don't need a mutable reference. This is however due to the computations done ahead of time that shouldn't be possible.

syldium (changes requested):
Handling captures requires modifying the current local layout to have relative locals. Pushing an intermediary reference of the local is a huge hack, and is not a proper capture system that must copy the value if the original variable goes out of scope.

analyzer/src/environment/variables.rs
```diff
@@ -31,7 +31,7 @@ pub struct Variables {
 
     /// Relations with external variables.
     /// The key is the variable Names, where value is the relation with another external environment's [Locals]
-    externals: HashMap<Name, RelationId>,
+    externals: IndexMap<Name, RelationId>,
```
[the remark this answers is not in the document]
Override-6: the compiler iterates over externals, we need a fixed order because otherwise the same source code would have different indexes for captured references

compiler/src/locals.rs
Override-6: it's not an error, if the index is already initialized for the given symbol, then no need to determine a new one. I'll refactor this to `init_external_ref_space` to be more explicit

analyzer/src/environment/variables.rs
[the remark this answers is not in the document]
syldium: As the captured list is recreated later by the compiler, this is not an issue.

Override-6:
Sorry for the misleading title, what this pull request really is about capturing stack values of outer functions, for example in the function
```kt
fun outer() = {
   var x = 1
   fun inner() = {
     $x = $x + 1
     echo $x
   }
   inner()
   inner()
}
```
`inner` will now be able to access to a reference of `x` localized in its outer function.
> Pushing an intermediary reference of the local is a huge hack, and is not a proper capture system that must copy the value if the original variable goes out of scope.
I understand what you mean but the current implementation of the analyzer does not supports clojures objects which currently makes impossible the writing of code that leaks a stack reference.

This pull request is about the capture by inner functions of its outer functions' values, which are the simplest cases of captures and thus a first reference-only implementation seems sufficient. I indeed admit that the title and pull request message was not exhaustive enough, changed the title

compiler/src/lib.rs
```diff
@@ -71,22 +69,109 @@ pub fn compile(
...
+
+        // recursively resolve all inner functions
+        for (_, func) in env
+            .variables
+            .iter()
+            .filter(|(_, v)| v.ty == TypeInfo::Function)
+        {
+            let func_fqn = &env.fqn.appended(Name::new(&func.name));
+            let (func_id, _) = engine.find_environment_by_name(func_fqn).unwrap();
+
```
syldium, line 101:
```suggestion
        for func_id in env.iter_direct_inner_environments() {
```
In `Environment`:
```rs
    /// Gets an iterator over the direct inner environment identifiers.
    pub fn iter_direct_inner_environments(&self) -> impl Iterator<Item = SourceId> + '_ {
        self.declarations.values().copied()
    }
```

compiler/src/lib.rs
```diff
@@ -71,22 +69,109 @@ pub fn compile(
...
+            }
+        });
+
+        captures[chunk_id.0] = Some(chunk_captures)
+    }
+
+    // resolve capture of all chunks, starting from root chunks of each module
+    for (chunk_id, _) in typed_engine
+        .iter_chunks()
+        .filter(|(id, _)| engine.get_environment(*id).unwrap().parent.is_none())
```
syldium, line 141:
```suggestion
        .filter(|(_, chunk)| chunk.is_script)
```

syldium (changes requested):

syldium (approved):

PR #130, opened 2023-07-23
Automatically test the VM implementation

syldium:
The VM is mostly tested manually or with primitive scripts. It's time to have a proper testing environment that directly targets the CLI binary.

This will allow playing with multiple files to testing the file importer in real conditions. For now, a few tests have been written with the [*lang_tester*](https://github.com/softdevteam/lang_tester/) crate that checks the execution output.

This PR also pull Criterion to benchmark the virtual machine performance. It has been added to a `vm` crate, that now contain the FFI code instead of the `cli` crate. The benchmarks belong to the crate directly, so compiling with the `test` profile will compile Criterion despite not being used. Benchmarks profile variable accesses and function calls.

### DIFF
--- a/vm/benches/exec_benchmark.rs
+++ b/vm/benches/exec_benchmark.rs
@@ -40,27 +40,31 @@ fn prepare_bytecode(code: &str) -> Vec<u8> {
 
 fn criterion_benchmark(c: &mut Criterion) {
     let bytes = prepare_bytecode(
-        "var u = 0
-    var computing = true
-    var v = 1
-    while $computing {
-        u = $u + 1
-        v = $v + 2
-        computing = $u != 20000
-    }",
+        "
+        var u = 0
+        var computing = true
+        var v = 1
+        while $computing {
+            u = $u + 1
+            v = $v + 2
+            computing = $u != 20000
+        }
+    ",
     );
     c.bench_function("var", |b| {
         b.iter(|| unsafe { execute_bytecode(black_box(&bytes)) })
     });
 
     let bytes = prepare_bytecode(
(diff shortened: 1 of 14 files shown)


vm/benches/exec_benchmark.rs
```diff
@@ -0,0 +1,71 @@
...
+    let bytes = prepare_bytecode(
+        "var u = 0
+    var computing = true
+    var v = 1
+    while $computing {
+        u = $u + 1
+        v = $v + 2
+        computing = $u != 20000
+    }",
+    );
```
Override-6, line 51:
```suggestion
    let bytes = prepare_bytecode("
        var u = 0
        var computing = true
        var v = 1
        while $computing {
            u = $u + 1
            v = $v + 2
            computing = $u != 20000
        }
    ",
    );
```

vm/benches/exec_benchmark.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    let bytes = prepare_bytecode(
+        "fun fibonacci(n: Int) -> Int =
+        if [ $n <= 1 ] {
+            1
+        } else {
+            return fibonacci($n - 1) + fibonacci($n - 2)
+        }
+    fibonacci(25)",
+    );
```
Override-6, line 64:
```suggestion
    let bytes = prepare_bytecode("
        fun fibonacci(n: Int) -> Int =
        if [ $n <= 1 ] {
            1
        } else {
            return fibonacci($n - 1) + fibonacci($n - 2)
        }
        fibonacci(25)
    ",
    );
```

Override-6 (approved):

PR #131, opened 2023-07-27
Load and store external symbols

syldium:
Support running files that import other files.

Bytecode format
==============

Single output
------------------

When compiling from the original sources, only one bytecode unit is really produced. The same constant pool is used for all the files, and all the main functions are concatenated, inserted between the functions of each file. The VM know when there are more functions to load by using the total length of the byte vector.

Multiple outputs
-------------------------

There are no possibilities to compile and run sparely currently, but this PR tries to set up it by allowing multiple constant pools to be loaded. This concept will be used very soon to persist the values used in the REPL.

Format
---------

The string constant pool now welcomes the names of the dynamic symbols to load and export. A dynamic symbol is identified by an identifier, and is itself a reference to the constant pool. This allows the dynamic symbols to be packed together to easily resolve them and to see if any of them is missing.

The exported symbols are written for each script chunk (as opposed to dynamic symbols), and points to the correct offset in a virtual memory page. The bytecode of the dependency may be able to change the offsets without requiring the user to recompile all the other bytecode units in a multiple files scenario.

Loader
=====

The VM now has responsibilities to link externals symbols before the actual program execution. It indexes the dynamic symbols to look for, then resolve them via a simple map lookup. Note that this system could be extended for user functions. Pages are memory sections that are kept alive during the whole program execution and are the central place where data is exchanged between the bytecode units.

Runner
======

The deepest dependencies are executed first, up to the entry point. Dynamic symbols directly access the memory pages. To some extent, it prefers safety by avoiding pointers that may be invalidated over time. Some usages can be changed, like instead of concatenating the instruction bytes, maybe create different vectors?

This PR don't mostly address name conflicts, where the same dynamic symbol is provided by multiple file. It is chosen arbitrary, at the moment.

Analyzer
=======

A small fix has been added to avoid recreating an environment if the same name is encountered multiple times (avoid compiling the same content multiple times!). A custom iterator is used to loop over the chunks with the same content id.


compiler/src/lib.rs
```diff
@@ -218,11 +233,37 @@ fn write(
...
+    for dynsym in &cp.dynsym {
+        writer.write_all(&dynsym.to_be_bytes())?;
+    }
+    Ok(())
+}
+
+fn write_exported(pool: &mut ConstantPool, writer: &mut impl Write) -> Result<(), io::Error> {
+    writer.write_all(
+        &u32::try_from(pool.exported.len())
+            .expect("too many locals")
```
Override-6, line 255:
```suggestion
            .expect("too many exported vars")
```

vm/src/definitions/loader.cpp
```diff
@@ -0,0 +1,96 @@
+#include "loader.h"
+
+#include "byte_reader.h"
+#include "conversions.h"
```
Override-6, line 4:
```suggestion
```

Override-6 (changes requested):

Override-6:
Also as you copies functions instructions, and as constant pool's strings are interned, you should free the bytecode once the unit is loaded.

Override-6 (approved):

PR #134, opened 2023-07-31
Add panic mechanism to VM, with optional source lines mappings.

Override-6:
This pull request adds a panic system to the runtime, with a call stack trace and optional source lines mappings. 

## Bytecode changes
Attributes were added to functions, `Code` attribute (identifier 1) contains the actual function definition and is required, the `Mappings` attribute (identifier 2) is optional and contains the mappings between function's instruction and source code lines.
Attributes can be read in any order.

## VM Implementation
The VM will print a panic message with its stack trace if any exception occurs inside the interpreter. To make the interpreter panic on purpose, for example if the VM is about to divide by zero, the new `RuntimeException` can be thrown with a message which will be the panic message.
```
panic: Attempted to divide 0 by zero.
    at sandbox.msh::with_captures::inner_with_captures::inner_with_captures_once (line 106)
    at sandbox.msh::with_captures::inner_with_captures (line 110)
    at sandbox.msh::with_captures (line 115)
    at sandbox.msh (line 129)
``` 

## Current limitation
No `panic` function is provided in this Pull request because current analyzer implementation cannot support predeclared symbols.


vm/src/memory/call_stack.cpp
```diff
@@ -115,3 +115,18 @@ size_t CallStack::size() const {
 bool CallStack::is_empty() const {
     return size() == 0;
 }
+std::vector<stack_frame_view> CallStack::dump_stack() const {
```
syldium, line 118: The stack must be unwound, i.e. all the frames should be removed. This reduces the need of using the `SIZE_MAX` constant trick, and would restore a sane VM state.

vm/src/interpreter.cpp
```diff
@@ -217,6 +228,26 @@ inline bool apply_comparison(Opcode code, double a, double b) {
     }
 }
 
+void panic_message(std::string msg, std::vector<stack_frame_view> stack) {
```
syldium, line 231:
```suggestion
void panic_message(const std::string &msg, std::span<const stack_frame> stack) {
```

syldium (changes requested):

compiler/src/bytecode.rs
```diff
@@ -91,13 +91,17 @@ pub struct Instructions<'a> {
     bytecode: &'a mut Bytecode,
     /// Offset of where this instruction set starts in the given bytecode
     ip_offset: u32,
+
+    /// sources segments with instructions ranges
```
syldium, line 95:
```suggestion
    /// Starting byte position in the source code for some instruction pointers
```

compiler/src/emit.rs
```diff
@@ -229,9 +230,23 @@ pub fn emit(
...
+            let var = ctx.environment.variables.get_var(d.identifier).unwrap();
+            if var.ty == TypeInfo::Function {
+                // if it's a function declaration, then do not bind any position with start position
+                // (no bytecode were emitted anyway)
+                return;
+            }
+        }
+        ExprKind::Block(exprs) => {
+            emit_block(exprs, instructions, ctx, cp, locals, state);
+            // do not bind generated bytecode with blocks' start positions
```
syldium, line 247: Is it an issue to still emit it for blocks and function declarations?

compiler/src/lib.rs
```diff
@@ -65,16 +75,79 @@ pub fn compile(
...
+    }
+}
+
+fn compile_line_mapping_attribute(
+    positions: Vec<(usize, u32)>,
+    content_id: ContentId,
+    bytecode: &mut Bytecode,
+    line_provider: &dyn SourceLineProvider,
+) {
+    // 2 is the mappings attribute identifier
```
syldium, line 124:
```suggestion
```

context/src/source.rs
```diff
@@ -3,6 +3,7 @@ use std::fmt::Debug;
 #[cfg(feature = "miette")]
 use miette::{MietteError, MietteSpanContents, SourceCode, SourceSpan, SpanContents};
 
+/// range of bytes of an unbound string source
```
syldium, line 6:
```suggestion
/// A range of bytes in an unbound string source.
```

vm/src/memory/call_stack.cpp
```diff
@@ -6,7 +6,7 @@
 struct frame_headers {
     /**
      * position of the previous frame header.
-     * 0 if this frame is the root headers
+     * SIZE_MAX if this frame is the root headers
```
syldium, line 9:
```suggestion
     * 0 if this frame is the root headers
```

vm/src/memory/call_stack.h
```diff
@@ -23,7 +28,7 @@ class CallStack {
 
     /**
      * position of the current frame headers
-     * is 0 if this call stack is empty
+     * is SIZE_MAX if this call stack is empty
```
syldium, line 31:
```suggestion
     * is 0 if this call stack is empty
```

vm/src/memory/call_stack.h
```diff
@@ -15,6 +15,11 @@ struct stack_frame {
     Locals locals;
 };
 
+struct stack_frame_view {
+    const function_definition &function;
+    size_t current_ip;
+};
+
```
syldium, line 22:
```suggestion
```

syldium (changes requested):
Since a division by zero will now not stop the REPL, the current redirections now also need to be cancelled: `shell { val m = 0 / 0 } > /dev/null`

compiler/src/lib.rs
```diff
@@ -65,16 +73,81 @@ pub fn compile(
...
+    line_provider: &dyn SourceLineProvider,
+) {
+    bytecode.emit_byte(MAPPINGS_ATTRIBUTE);
+    let mut mappings: Vec<(usize, u32)> = Vec::new();
+
+    let Some(((mut last_pos, mut last_ip), positions)) = positions.split_first() else {
+        bytecode.emit_u32(0);
+        return
+    };
+    let mut last_line = usize::MAX;
```
syldium, line 129: The initial position is never inserted.

syldium (changes requested):
```kt
fun foo() -> Int =
    5
    /
    0

fun bar() -> ExitCode =
    echo \
    > \ // Expected line
    /not

bar() // Actual line: 3
```
The lines are incorrect under specific conditions.

Override-6:
> ```kotlin
> fun foo() -> Int =
>     5
>     / // Actual line
>     0
> 
> echo \
> > \ // Expected line
> /not
> ```
> 
> The lines are incorrect under specific conditions.

the `foo` function actually returns `5`, and the `/` is actually seen as a command on script's level. it attempts to run the  `/` cmd which causes a panic. Made more elabored panic messages :
```
panic: Unable to execute command "/": Permission denied
	at sandbox.msh (line 3)
```

compiler/src/lib.rs
Override-6: it is expected to be masked out when there are many positions, i added a case to effectively append the first pos' line to the first instruction (#0) if `mappings` is empty

compiler/src/lib.rs
```diff
@@ -22,59 +23,137 @@ mod r#type;
...
+            continue;
+        }
+        last_pos = pos;
+    }
+
+    if mappings.is_empty() {
+        // if no mappings are set, bind first pos' line with first instruction.
+        let line = line_provider.get_line(content_id, *first_pos).unwrap();
+        mappings.push((line, 0))
+    }
```
syldium, line 148: Given that this function has gradually increased its complexity, it could be better to split the responsibilities. Make that the `Instructions::push_position` method don't re-push on the same start segment to de-duplicate ahead of time.

vm/src/interpreter.cpp
```diff
@@ -389,8 +441,13 @@ bool run_frame(runtime_state &state, stack_frame &frame, CallStack &call_stack,
...
             // Replace the current process with a new process image
             if (execvp(argv[0].get(), reinterpret_cast<char *const *>(argv.data())) == -1) {
-                perror("execvp");
-                _exit(MOSHELL_COMMAND_NOT_RUNNABLE);
+                std::stringstream ss;
+                ss << argv[0].get();
+                for (int i = 1; i < frame_size; i++) {
+                    ss << " " << std::string(argv[i].get());
+                }
+                panic("Unable to execute command \"" + ss.str() + "\": " + std::string(strerror(errno)), call_stack);
```
syldium, line 449:
```suggestion
                std::string command = argv[0].get();
                panic("Unable to execute command \"" + command + "\": " + std::string(strerror(errno)), call_stack);
```
An `ostream` is not particularly needed here, I am not very convinced that the full command should be printed.

vm/src/interpreter.cpp
```diff
@@ -355,9 +406,10 @@ bool run_frame(runtime_state &state, stack_frame &frame, CallStack &call_stack,
             pid_t pid = fork();
             switch (pid) {
             case -1:
-                perror("fork");
-                exit(EX_OSERR);
+                panic(strerror(errno), call_stack);
+                return frame_status::ABORT;
             case 0:
+                is_master = false;
                 // Child process
```
syldium, line 413:
```suggestion
                // Child process
                is_master = false;
```

vm/src/interpreter.cpp
```diff
@@ -217,6 +233,41 @@ inline bool apply_comparison(Opcode code, double a, double b) {
...
+                if (instruction >= *frame.instruction_pointer) {
+                    break;
+                }
+                instruction_line = line;
+            }
+            std::cerr << " (line " << instruction_line << ")";
+        }
+        std::cerr << std::endl;
+        stack.pop_frame();
+    }
```
syldium, line 256:
```suggestion
    std::cerr << "panic: " << msg;

    while (!stack.is_empty()) {
        stack_frame frame = stack.peek_frame();
        const function_definition &def = frame.function;
        std::cerr << "\tat " << *def.identifier;

        if (!def.mappings.empty()) {
            size_t instruction_line;
            for (const auto &[instruction, line] : def.mappings) {
                if (instruction >= *frame.instruction_pointer) {
                    break;
                }
                instruction_line = line;
            }
            std::cerr << " (line " << instruction_line << ")";
        }
        std::cerr << "\n";
        stack.pop_frame();
    }
    std::cerr << std::endl;
```
Flushes only once.

syldium (changes requested):

compiler/src/lib.rs
Override-6: i'd like to let the `Instructions::push_position` as simple as possible and make all the computation if the mappings are required

vm/src/interpreter.h
```diff
@@ -9,9 +9,18 @@ namespace msh {
...
 }
 
+class RuntimeException : std::exception {
+private:
+    const std::string msg;
+
+public:
+    explicit RuntimeException(std::string msg);
+    [[nodiscard]] const char *what() const noexcept override;
+};
```
syldium, line 19:
```suggestion
class RuntimeException : public std::runtime_error {
public:
    explicit RuntimeException(std::string msg);
};
```

vm/src/interpreter.cpp
```diff
@@ -217,6 +233,53 @@ inline bool apply_comparison(Opcode code, double a, double b) {
     }
 }
 
+void panic(const std::string& msg, CallStack &stack) {
```
syldium, line 236:
```suggestion
void panic(const std::string &msg, CallStack &stack) {
```

vm/src/interpreter.cpp
```diff
@@ -123,6 +128,11 @@ struct runtime_state {
     const natives_functions_t &native_functions;
 };
 
+RuntimeException::RuntimeException(std::string msg) : std::exception(), msg{msg} {}
+const char *RuntimeException::what() const noexcept {
+    return msg.c_str();
+}
```
syldium, line 134:
```suggestion
RuntimeException::RuntimeException(std::string msg)
    : std::runtime_error(std::move(msg)) {}
```

vm/src/interpreter.cpp
```diff
@@ -14,9 +14,9 @@
 #include <cstring>
 #include <fcntl.h>
 #include <iostream>
+#include <sstream>
```
syldium, line 17:
```suggestion
```

vm/src/interpreter.cpp
```diff
@@ -217,6 +233,53 @@ inline bool apply_comparison(Opcode code, double a, double b) {
     }
 }
 
+void panic(const std::string& msg, CallStack &stack) {
+    std::cerr << "panic: " << msg;
```
syldium, line 237:
```suggestion
    std::cerr << "panic: " << msg << "\n";
```

syldium (changes requested):
Please stick to one code style.

compiler/src/lib.rs
syldium: The tradeoff doesn't seem fair, as it increases code complexity here and increases memory usage as a lot is duplicated.

syldium (approved):

PR #137, opened 2023-08-12
remove inclusion path and type arguments from command calls

Override-6:
This pull request removes from the AST the inclusion `path` and `type_parameters` fields.
As we discussed that moshell functions and system command calls cannot get mixed up, type parameters and inclusion paths makes no more sense in the current moshell's state

### DIFF
--- a/parser/src/aspects/call.rs
+++ b/parser/src/aspects/call.rs
@@ -59,7 +59,7 @@ impl<'a> CallAspect<'a> for Parser<'a> {
             )
             .is_none()
         {
-            check_emptiness("Command calls cannot have inclusion path", path, |str| {
+            ensure_empty("Command calls cannot have inclusion path", path, |str| {
                 self.cursor.relative_pos(*str)
             })?;
             return self.call();
@@ -91,12 +91,12 @@ impl<'a> CallAspect<'a> for Parser<'a> {
             .advance(spaces().then(of_type(TokenType::FatArrow)))
             .is_some()
         {
-            check_emptiness(
+            ensure_empty(
                 "Illegal expression before lambda parameter declaration",
                 path,
                 |str| self.cursor.relative_pos(*str),
             )?;
-            check_emptiness(
+            ensure_empty(
                 "Illegal expression after lambda parameter declaration",
                 type_parameters,
                 Type::segment,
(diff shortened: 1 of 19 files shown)


parser/src/aspects/type.rs
```diff
@@ -78,11 +78,7 @@ impl<'a> TypeAspect<'a> for Parser<'a> {
     }
 
     fn parse_type_parameter_list(&mut self) -> ParseResult<(Vec<Type<'a>>, SourceSegment)> {
-        if self
-            .cursor
-            .lookahead(blanks().then(of_type(SquaredLeftBracket)))
-            .is_none()
-        {
+        if self.cursor.lookahead(of_type(SquaredLeftBracket)).is_none() {
```
syldium, line 81: Since this parser is also used for function declarations, it should be noted that `fun foo [T]() = {}` is now an invalid input if not updated.

parser/src/aspects/call.rs
```diff
@@ -224,13 +233,28 @@ impl<'a> CallAspect<'a> for Parser<'a> {
...
 }
 
-impl<'a> Parser<'a> {
-    fn call_with_path(&mut self, path: Vec<&'a str>) -> ParseResult<Expr<'a>> {
-        let callee = self.call_argument()?;
-        let tparams = self.parse_type_parameter_list()?.0;
-        self.call_arguments(path, callee, tparams)
+fn check_emptiness<T>(
+    msg: &str,
+    elements: Vec<T>,
```
syldium, line 238: I see a world where it makes sense to take ownership of the data to ensure that it will not be used later. It would be nice to document that, as the name of the function does not suggest that. Maybe rename it to `ensure_empty` or `fail_if_some`?

syldium (changes requested):

parser/src/aspects/type.rs
Override-6: I think that the fact main parser methods will automatically remove blanks / spaces at the begining is a conception issue that should be removed in a later PR, I fixed it here 'cause i needed more control over heading blanks in `any_call` but indeed i haven't anticipated the  change needed to by other methods to remove heading blanks before calling it.

syldium (approved):

PR #139, opened 2023-08-16
Introduce reefs on parser side

Override-6:
This pull request introduces a way to distinct a source file's current and external reef symbols.


parser/src/aspects/call.rs
```diff
@@ -226,36 +240,17 @@ impl<'a> CallAspect<'a> for Parser<'a> {
     fn may_be_at_programmatic_call_start(&self) -> bool {
         self.cursor
             .lookahead(
-                repeat(of_type(Identifier).and_then(of_type(ColonColon)))
+                of_type(TokenType::Reef)
+                    .and_then(of_type(TokenType::ColonColon))
+                    .then(repeat(
+                        of_type(TokenType::Identifier).and_then(of_type(TokenType::ColonColon)),
+                    ))
```
syldium, line 247:
```suggestion
                repeat(
                    of_types(&[TokenType::Identifier, TokenType::Reef])
                        .and_then(of_type(TokenType::ColonColon)),
                )
```
Let the actual path validator decides if the path is actually valid.

parser/src/parser.rs
```diff
@@ -254,7 +254,9 @@ impl<'a> Parser<'a> {
             //expression that can also be used as values
             If => self.parse_if(Parser::value).map(Expr::If),
             Match => self.parse_match(Parser::value).map(Expr::Match),
-            Identifier if self.may_be_at_programmatic_call_start() => self.programmatic_call(),
+            Identifier | Reef if self.may_be_at_programmatic_call_start() => {
```
syldium, line 257: The `call_argument` method should also be updated.

syldium (changes requested):
`reef::reef()` is currently a valid function call.

parser/src/aspects/call.rs
```diff
@@ -75,16 +78,27 @@ impl<'a> CallAspect<'a> for Parser<'a> {
         let type_parameters = self.parse_type_parameter_list()?.0;
         if let Some(open_parenthesis) = self.cursor.advance(of_type(TokenType::RoundedLeftBracket))
         {
+            if identifier.token_type != Identifier {
+                return self.expected(
+                    "Expected function name.",
+                    ParseErrorKind::Expected("valid function identifier".to_string()),
+                );
```
syldium, line 85:
```suggestion
                return self.expected_with(
                    "Expected function name.",
                    identifier.clone(),
                    ParseErrorKind::UnexpectedInContext(format!(
                        "`{}` is a reserved keyword.",
                        identifier.value
                    )),
                );
```
This error is placed one token after the erroneous token, and doesn't state that the token is a keyword. I also have doubts about the placement of this code, as keywords are a more complex issue below the surface (see `echo while()`, `echo reef()` vs `reef()` behave quite differently).

syldium (changes requested):

analyzer/src/types/ctx.rs
```diff
@@ -70,10 +71,11 @@ impl TypeContext {
     pub(crate) fn resolve(&self, type_annotation: &ast::r#type::Type) -> Option<TypeId> {
         match type_annotation {
             ast::r#type::Type::Parametrized(param) => {
-                if !param.path.is_empty() || !param.params.is_empty() {
+                if param.path.len() > 1 || !param.params.is_empty() {
                     unimplemented!();
                 }
-                self.names.get(param.name).copied()
+                let InclusionPathItem::Symbol(sym, _) = param.path.first().unwrap() else { unimplemented!() };
```
syldium, line 77:
```suggestion
                let InclusionPathItem::Symbol(sym, _) = param.path.first().expect("Type annotation should not be empty") else {
                    return None; // Type annotations may only be symbols
                };
```

parser/src/aspects/call.rs
```diff
@@ -161,14 +196,19 @@ impl<'a> CallAspect<'a> for Parser<'a> {
...
-            )
-        } else {
-            None
-        };
+                    .force(of_type(TokenType::Identifier), "Expected function name.")
+                    .map_err(|err| {
+                        err.with_kind(ParseErrorKind::Expected(
+                            "valid function identifier".to_string(),
+                        ))
+                    })
```
syldium, line 208:
```rs
                self.cursor.force_with(
                    of_type(TokenType::Identifier),
                    "Expected function name.",
                    ParseErrorKind::Expected("identifier".to_owned()),
                )
```

parser/src/aspects/call.rs
```diff
@@ -46,45 +46,64 @@ pub trait CallAspect<'a> {
...
+            segment: name_segment.clone(),
         });
         let type_parameters = self.parse_type_parameter_list()?.0;
         if let Some(open_parenthesis) = self.cursor.advance(of_type(TokenType::RoundedLeftBracket))
         {
+            if identifier.token_type != TokenType::Identifier {
+                return self.expected_with(
+                    "Expected function name.",
+                    identifier,
+                    ParseErrorKind::Expected("valid function identifier".to_string()),
```
syldium, line 89:
```suggestion
                    ParseErrorKind::Expected("identifier".to_owned()),
```
All identifiers are valid.

syldium (changes requested):

syldium (approved):

PR #141, opened 2023-09-03
Introduce vectors

syldium:
Vectors are growable arrays that can contain an arbitrary number of items that are indexed by an unsigned integer. The generic type `Vec` is now known by the analyzer. If everything is fine, it is passed to the compiler and the virtual machine, that boxes primitives.

Representing generics
--------------------------------

The `Type` enum could already represent user defined functions, but in a limited way. The parameter types could not be displayed properly, and the equality relationship between two `TypeId`s can be false when the two `Type`s are identical. The newly added `TypeInstance` struct has a way to print the entire type, and `Typing` is used to test if two types are compatible. It knows if a generic type is concretized in the current context.

Poisoned expressions
-------------------------------

The typing phase now prefers to poison types in case of errors. For instance, `val v: Vec[Int] = 4` now prefer using the type annotation instead of the actual type for the variable type. It sometimes reduces the number of errors.

Type annotations
-------------------------

The type annotation collection will move outside the typing phase, but it is still important to ensure that generic parameters are valid. Some helpers functions have been added to ensure that they are properly checked.

Virtual machine
----------------------

This PR also adds a simple runtime implementation for vectors, that needs to box primitives.


compiler/src/emit/native.rs
```diff
@@ -1,24 +1,32 @@
...
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn emit_natives(
     native: NativeId,
-    callee: &TypedExpr,
-    args: &[TypedExpr],
+    MethodCall {
+        callee,
+        arguments: args,
+        ..
+    }: &MethodCall,
```
Override-6, line 28: only `callee` is used
```suggestion
    callee: &TypedExpr,
```

compiler/src/emit.rs
```diff
@@ -289,3 +285,8 @@ pub fn emit(
     }
     instructions.push_position(expr.segment.start)
 }
+
+/// Tests if a type is a primitive type that needs to be boxed or unboxed.
+fn is_primitive(ty: &Type) -> bool {
```
Override-6, line 290:
```suggestion
fn is_boxable_primitive(ty: &Type) -> bool {
```

Override-6 (changes requested):

compiler/src/emit/native.rs
syldium: `args` is used every time a parameter need to be emitted.

Override-6: damn

Override-6 (approved):

PR #142, opened 2023-09-03
Implement reefs on analyzer side

Override-6:
This pull request follows the previous parser side pull request (#139) to implement reefs on the analyzer.

A reef is a structure that contains the analyzer's outputs over a set of input sources (`Engine`, `Relations`, `TypedEngine`, `Typing` and `TypeContext`).
A reef can also use symbols from other reefs, which needed the resolver, collector and `Relations` to take some significant changes that'll be enumerated later.

In order for this pull request to make sense, i also made types behave like regular symbols, and now the collection/resolution phases also works for types ! which also implied many changes on the typing phase.

## `Reef` / `Reefs`
the `Reef` structure contains all the symbols, types and relations of a reef, with a given name. 
Available reefs are stored into a `Reefs` structure, and reef's external relations are relative to it (for example, a reef pointing to the reef with id 4 is the reef with id 4 of its own `Reefs` container.)

## `ReefContext` / collection / resolution / typing phases

The ReefContext is a structure that mutably borrows a given `Reefs` structure, and reserves a reef in it. The reserved reef can then be mutably accessed using `ReefContext::current_reef_mut`. The `reefs` structure is protected by the context and is only exposed immutably.

Now that the analyzer's data is stored into a `Reef` structure, the `Engine` and `Relations` where removed from the `ResolutionResult` (used by collect/resolution phases).

But the fact that all the data is now contained in a `Reefs` structure does not really enjoys the borrow checker...  

In the collection / resolution phase this is not a real concern, and a smart use of `ReefContext::current_reef_mut` was sufficient to resolve the borrow issues.  
But in the `typing` phase, because some immutable access to the external reefs needed to be done even if the current reef's types structures are modified, the `Typing`, `TypeContext` and `TypedEngine` structures needed to still be owned by the `Exploration` structure.
This makes the access to a reef's type more complicated because the current reef's types cannot be accessed through the `Reefs` as they are written at the end of the typing phase from the `Exploration` structure. 

### URA
To overcome this issue, the `UniversaReefAccessor` is used by most of the typing phase functions, which is a proxy that takes the current exploration's types data (`Typing`, `TypeContext` and `TypedEngine`) and returns under the same interface the exploration's type data or any other reef's data if the given reef identifier is the current reef's identifier

## SymbolRegistry 
As types are now handled as regular symbols, but types can also take the same name of a variable or a function's name, the symbols now lives in a defined registry, which is either `Types` for types, or `Objects` for variables and functions.

## Other changes
Refactors :
- `environment::variables::Variables` is now `environment::symbols::Symbols`,
- The `relations::Symbol` enum became `relations::SymbolRef`
- The `Variable` structure in `environment::variables` became `Symbol`

The monstruous `ExitCode` type has been renamed with the fancy `Exitcode` name, yes this is arbitrary lol but we can discuss if you want for a better name, because i personally really dislike the way it's currently named.

### DIFF
--- a/analyzer/src/steps/typing/exploration.rs
+++ b/analyzer/src/steps/typing/exploration.rs
@@ -1,5 +1,5 @@
 use crate::engine::Engine;
-use crate::reef::{ReefAccessor, ReefId, Reefs};
+use crate::reef::{ReefId, Reefs};
 use crate::relations::Relations;
 use crate::steps::typing::function::Return;
 use crate::types::ctx::TypeContext;
@@ -21,6 +21,9 @@ pub struct ReefTypes<'a> {
     pub typing: &'a Typing,
 }
 
+/// A proxy that let the user access to any reef's engine, relations and types data.
+/// If the given reef identifier is the `current_reef`, the types data stored in `current` is returned instead
+/// of the data stored by the `reefs`.
 pub struct UniversalReefAccessor<'a, 'e> {
     reefs: &'a Reefs<'e>,
     current_reef: ReefId,
@@ -54,6 +57,8 @@ impl Exploration {
         self.returns.clear();
     }
 
+    /// returns an universal accessor that will return the exploration's types data
+    /// if the accessed reef's identifier is equal to given `current_reef`
     pub(crate) fn universal_accessor<'a, 'e>(
         &'e self,
         current_reef: ReefId,
--- a/analyzer/src/steps/typing/exploration.rs
+++ b/analyzer/src/steps/typing/exploration.rs
@@ -14,7 +14,7 @@ pub(super) struct Exploration {
     pub(super) returns: Vec<Return>,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Copy)]
 pub struct ReefTypes<'a> {
     pub context: &'a TypeContext,
     pub engine: &'a TypedEngine,
@@ -24,6 +24,7 @@ pub struct ReefTypes<'a> {
 /// A proxy that let the user access to any reef's engine, relations and types data.
 /// If the given reef identifier is the `current_reef`, the types data stored in `current` is returned instead
 /// of the data stored by the `reefs`.
+#[derive(Clone, Copy)]
 pub struct UniversalReefAccessor<'a, 'e> {
     reefs: &'a Reefs<'e>,
     current_reef: ReefId,
@@ -33,7 +34,7 @@ pub struct UniversalReefAccessor<'a, 'e> {
 impl<'a, 'e> UniversalReefAccessor<'a, 'e> {
     pub fn get_types(&self, id: ReefId) -> Option<ReefTypes<'a>> {
         if id == self.current_reef {
-            Some(self.current.clone())
+            Some(self.current)
         } else {
             self.reefs.get_reef(id).map(|reef| ReefTypes {
                 context: &reef.type_context,
--- a/analyzer/src/steps/typing/exploration.rs
+++ b/analyzer/src/steps/typing/exploration.rs
@@ -1,20 +1,52 @@
-use context::source::SourceSegment;
-
-use crate::diagnostic::{Diagnostic, DiagnosticID, Observation};
-use crate::relations::SourceId;
+use crate::engine::Engine;
+use crate::reef::{ReefAccessor, ReefId, Reefs};
+use crate::relations::Relations;
 use crate::steps::typing::function::Return;
 use crate::types::ctx::TypeContext;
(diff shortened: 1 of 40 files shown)


analyzer/src/steps/typing/exploration.rs
```diff
@@ -1,34 +1,77 @@
...
+pub struct ReefTypes<'a> {
+    pub context: &'a TypeContext,
+    pub engine: &'a TypedEngine,
+    pub typing: &'a Typing,
+}
+
+/// A proxy that let the user access to any reef's engine, relations and types data.
+/// If the given reef identifier is the `current_reef`, the types data stored in `current` is returned instead
+/// of the data stored by the `reefs`.
+pub struct UniversalReefAccessor<'a, 'e> {
```
syldium, line 27: I have doubts about this structure that only allow immutable access. If it is really immutable, mark it `Clone` and `Copy`.

My concern here is that it impose immutable access everywhere, where the current `Exploration` is mutable.

analyzer/src/types.rs
```diff
@@ -17,97 +21,159 @@ pub struct Typing {
...
+}
+
+pub fn get_type<'a>(type_ref: TypeRef, ura: &'a UniversalReefAccessor) -> Option<&'a Type> {
+    ura.get_types(type_ref.reef)
+        .and_then(|types| types.typing.get_type(type_ref.type_id))
+}
+
+/// Finds the type reference from an annotation.
+pub(crate) fn resolve_type(
+    ura: &UniversalReefAccessor,
```
syldium, line 120: If a call concretize a generic type in #141, it needs to instantiate a new type in the `TypeContext`. `UniversalReefAccessor` is passed as a reference, so it could become a mutable reference, but everything behind is immutable references. Maybe use a mutable reference to the `Exploration` that could hold the immutable references to the other reefs?

analyzer/src/types/builtin.rs
```diff
@@ -74,28 +81,76 @@ pub fn lang(engine: &mut TypedEngine) {
...
+    }
+    typing.set_implicit_conversion(EXIT_CODE.type_id, BOOL);
+    typing.set_implicit_conversion(INT.type_id, FLOAT);
+}
+
+fn fill_lang_bindings(ctx: &mut TypeContext) {
+    ctx.bind_name("Nothing".to_string(), NOTHING.type_id);
+    ctx.bind_name("Unit".to_string(), UNIT.type_id);
+    ctx.bind_name("Bool".to_string(), BOOL.type_id);
+    ctx.bind_name("Exitcode".to_string(), EXIT_CODE.type_id);
```
syldium, line 135: If `ExitCode` is one word, `EXIT_CODE` becomes `EXITCODE`.

cli/src/repl.rs
```diff
@@ -35,6 +41,7 @@ pub fn prompt(mut importer: FileImporter, config: &Cli) -> PipelineStatus {
                     name: name.clone(),
                     imported,
                     attached: starting_source,
+                    reef: reef_id,
```
syldium, line 44: This field is never taken into account when injecting.

analyzer/src/types/ctx.rs
```diff
@@ -67,26 +51,12 @@ impl TypeContext {
...
-                }
-            }
-            ast::r#type::Type::Callable(_) => unimplemented!(),
-            ast::r#type::Type::ByName(_) => unimplemented!(),
-        }
+    pub(crate) fn bind_name(&mut self, name: String, tpe: TypeId) {
+        self.names.insert(name, tpe);
+    }
+
+    pub fn get_type(&self, name: &str) -> Option<TypeId> {
```
syldium, line 58:
```suggestion
    pub fn get_type_id(&self, name: &str) -> Option<TypeId> {
```

analyzer/src/types.rs
```diff
@@ -17,97 +21,159 @@ pub struct Typing {
...
+) -> TypeRef {
+    match type_annotation {
+        ast::r#type::Type::Parametrized(param) => {
+            if param.path.len() > 1 || !param.params.is_empty() {
+                unimplemented!();
+            }
+            if let InclusionPathItem::Symbol(_, _) = param
+                .path
+                .first()
+                .expect("Type annotation should not be empty")
```
syldium, line 133: Any reason to perform this check since the symbol is not used as it is already resolved?

syldium (changes requested):

analyzer/src/types.rs
Override-6: no it's just an oversight

syldium (approved):

PR #148, opened 2023-09-16
Bootstrap Garbage Collector

Override-6:
Adds a simple garbage collector to the Virtual Machine.
Virtual Objects now have a byte header that is used by the GC to mark its current cycle.

For each runs, the GC searches for the root objects (exported variables, constants and callstack), then does a in-depth course and marks the objects with its current cycle number.
Then it iterates of the vm's heap and removes all objects that does not have the GC's cycle number.

In order for the GC to locate object references in heap, this PR also adds a object references offsets array for functions locals, and a runtime list to register operands that are object references.


vm/src/memory/call_stack.cpp
```diff
@@ -119,4 +126,47 @@ bool CallStack::is_empty() const {
...
+call_stack_iterator CallStack::begin() {
+    return call_stack_iterator(this, frame_headers_pos, size());
+}
+
+call_stack_iterator CallStack::back() {
+    if (is_empty())
+        return end();
+    return call_stack_iterator(this, 0, 1);
+}
+
```
syldium, line 140:
```suggestion
```
`back()` returns a reference to the last element in the stack, not an iterator. Given that `front()` does not exist, it is a strange addition here.

vm/src/memory/heap.h
```diff
@@ -13,15 +13,42 @@ namespace msh {
...
+     */
+    class obj {
+        uint8_t gc_cycle;
+        obj_data data;
+
+        friend gc;
+
+    public:
+        template <typename T>
+        obj(T val) : gc_cycle{0}, data{val} {}
```
syldium, line 38:
```suggestion
        obj(obj_data val) : gc_cycle{}, data{std::move(val)} {}
```
A bit opinionated, but `std::variant` already has a non-explicit constructor. It does break code in `constant_pool.cpp` through.

vm/src/stdlib_natives.cpp
```diff
@@ -177,7 +187,9 @@ natives_functions_t load_natives(msh::heap &) {
         {"std::set_env", set_env},
         {"std::is_env_def", is_env_def},
         {"std::read_line", read_line},
-        
+
+        {"std::intern::gc", gc},
```
syldium, line 191:
```suggestion
        {"std::memory::gc", gc},
```
Intern is for strings.

vm/src/memory/operand_stack.h
```diff
@@ -91,20 +94,34 @@ class OperandStack {
...
     void push(const char *bytes, size_t size);
 
     template <typename T>
     void push(T t) {
         if (current_pos + sizeof(T) > stack_capacity) {
             throw StackOverflowError("exceeded stack capacity via operand stack");
         }
+
+        size_t false_bits_count = sizeof(T);
+        if constexpr (std::is_same_v<T, msh::obj *>) {
```
syldium, line 113: This test is incorrect when using `const msh::obj*`, `msh::obj const*`, `msh::obj&`...

vm/src/memory/operand_stack.cpp
```diff
@@ -54,14 +55,26 @@ const char *OperandStack::pop_bytes(size_t n) {
...
     }
+    operands_refs.resize(operands_refs.size() - n);
     current_pos -= n;
     return bytes + current_pos;
 }
 
+void OperandStack::transfer(OperandStack &caller_stack, size_t n) {
+#ifndef NDEBUG
+    if (n > caller_stack.size())
+        throw std::out_of_range("cannot transfer more byte than contained in source operand stack");
```
syldium, line 66:
```suggestion
    if (n > caller_stack.size()) {
        throw std::out_of_range("cannot transfer more bytes than contained in the source operand stack");
    }
```

vm/src/memory/operand_stack.cpp
```diff
@@ -54,14 +55,26 @@ const char *OperandStack::pop_bytes(size_t n) {
...
+#endif
+    memcpy(this->bytes + current_pos, caller_stack.bytes + (caller_stack.size() - n), n);
```
syldium, line 68: Given that this function could accept the same instance, it would be great to document to not to. `memcpy` doesn't play well if the two pointers point to overlapping memory regions.

vm/src/memory/gc.cpp
```diff
@@ -0,0 +1,187 @@
...
+    std::vector<bool> &operands_refs_offsets = thread_call_stack.operands_refs_offsets;
+
+    // the offsets will be scanned backward (following frames stack iteration)
+    size_t operands_bytes_index = operands_refs_offsets.size();
+
+    for (stack_frame frame : thread_call_stack) {
+        const Locals locals = frame.locals;
+        // add locals roots
+        for (size_t obj_ref_offset : frame.function.obj_ref_offsets) {
+            msh::obj *obj_ref = (msh::obj *)locals.get_q_word(obj_ref_offset);
```
syldium, line 145: Don't use a C-style cast here, define a proper method to get a pointer to restricted types in the locals.

vm/src/memory/gc.cpp
```diff
@@ -0,0 +1,187 @@
...
+}
+
+void gc::scan_thread(std::vector<msh::obj *> &roots) {
+    std::vector<bool> &operands_refs_offsets = thread_call_stack.operands_refs_offsets;
+
+    // the offsets will be scanned backward (following frames stack iteration)
+    size_t operands_bytes_index = operands_refs_offsets.size();
+
+    for (stack_frame frame : thread_call_stack) {
+        const Locals locals = frame.locals;
```
syldium, line 142:
```suggestion
        const Locals &locals = frame.locals;
```

vm/src/memory/gc.cpp
```diff
@@ -0,0 +1,187 @@
...
+        } else if constexpr (std::is_same_v<T, int64_t>) {
+            debug_file << "boxed int64 " << data;
+        } else if constexpr (std::is_same_v<T, double>) {
+            debug_file << "boxed double " << data;
+        } else if constexpr (std::is_same_v<T, obj_vector>) {
+            obj_vector &vec = static_cast<obj_vector &>(data);
+            debug_file << "vector len:" << vec.size();
+        } else {
+            // unreachable
+            throw MemoryError("unknown object type");
```
syldium, line 52:
```suggestion
            static_assert(false, "non-exhaustive object visitor!");
```

syldium (changes requested):

vm/src/memory/operand_stack.h
```diff
@@ -91,20 +94,36 @@ class OperandStack {
...
     void push(const char *bytes, size_t size);
 
     template <typename T>
     void push(T t) {
         if (current_pos + sizeof(T) > stack_capacity) {
             throw StackOverflowError("exceeded stack capacity via operand stack");
         }
+
+        size_t false_bits_count = sizeof(T);
+        if constexpr ((std::is_pointer_v<T> || std::is_reference_v<T>) && std::is_same_v<std::decay_t<T>, msh::obj*>) {
```
syldium, line 115:
```suggestion
        if constexpr (std::is_same_v<std::remove_cvref_t<std::remove_pointer_t<T>>, msh::obj>) {
```

syldium (changes requested):

syldium (approved):
Really cool! (don't forget the unresolved comment)

PR #152, opened 2023-09-19
Add functional tests for the virtual machine

Override-6:
This pull request adds more foreign function interfaces to the VM to have more control about its gc and manipulate its values / object.
A `Runner` utility has been created in the tests crate of the VM to evaluate expressions, invoke gc and manipulated gced objects and objects returned from evaluations


analyzer/src/lib.rs
```diff
@@ -135,6 +136,14 @@ impl<'a> Analyzer<'a> {
...
             &mut visit,
         ));
+        self.diagnostics.extend(SymbolResolver::resolve_symbols(
+            &self.resolution.engine,
+            &mut self.resolution.relations,
+            &mut self.resolution.imports,
+            externals,
+            &mut visit,
+            &self.resolution.visited,
+        ));
```
syldium, line 146: This is a regression, the REPL must be able to import other files.

cli/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub mod pipeline;
+
+pub mod disassemble;
+
+pub mod cli;
+
+mod report;
```
syldium, line 7:
```suggestion
mod cli;
mod disassemble;
pub mod pipeline;
mod report;
```

compiler/src/lib.rs
```diff
@@ -325,6 +344,26 @@ fn compile_chunk_code(
         &mut state,
     );
 
+    let chunk_is_script = ctx.environment.is_script;
+
+    if let Some(storage_exported_val) = &options.last_page_storage_var {
+        debug_assert!(chunk_is_script); // only script chunks can store their last expression value in a storage export
```
syldium, line 350:
```suggestion
        assert!(
            chunk_is_script,
            "only script chunks can store their last expression value in a storage export"
        );
```

lib/std/convert.msh
```diff
@@ -2,6 +2,9 @@
 /// @panics if i cannot be converted to an exitcode, i.e it is negative or superior to 256
 fun to_exitcode(i: Int) -> Exitcode;
 
+/// converts given exitcode to an integer
+fun to_int(e: Exitcode) -> Int;
+
```
syldium, line 7: It is a method on `ExitCode` or a cast.

parser/tests/expr.rs
```diff
@@ -21,6 +21,8 @@ use context::str_find::{find_between, find_in, find_in_nth};
 use parser::parse;
 use parser::source::{literal, literal_nth};
 
+pub struct Wtf {}
+
```
syldium, line 25:
```suggestion
```

vm/binding/lib.rs
```diff
@@ -1,6 +1,16 @@
-use context::source::ContentId;
+#![allow(unused, dead_code, warnings)]
```
syldium, line 1:
```suggestion
```
Address the warnings, they are important.

vm/binding/lib.rs
```diff
@@ -69,6 +83,32 @@ impl VM {
...
+            let size = result.collected_objects_count as usize;
+            let mut vec = Vec::with_capacity(size);
+
+            for i in 0..size {
+                let obj = *(result.collected_objects.wrapping_add(i));
+                vec.push(VmValue::from(obj))
+            }
+            vec
+        }
+    }
```
syldium, line 111: Do a separate file for that.

vm/src/definitions/pager.h
```diff
@@ -96,6 +96,12 @@ namespace msh {
 
         page_vector::const_reverse_iterator cend() const;
 
+        template <typename T>
+        T *get_exported_value(exported_variable var) {
```
syldium, line 100: Return a `moshell_value` type.

vm/src/vm.h
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <cstdint>
```
syldium, line 2:
```suggestion
#include <stdint.h> // NOLINT(*-deprecated-headers)
```
This is a C compatible header file.

vm/src/vm.h
```diff
@@ -14,10 +15,43 @@
...
+} moshell_array;
+
+uint8_t moshell_value_get_as_byte(moshell_value val);
+int64_t moshell_value_get_as_int(moshell_value val);
+double moshell_value_get_as_double(moshell_value val);
+moshell_object moshell_value_get_as_object(moshell_value val);
+
+moshell_value moshell_object_unbox(moshell_object obj);
+const char *moshell_object_get_as_string(moshell_object obj);
+moshell_array moshell_object_get_as_array(moshell_object obj);
```
syldium, line 49: Please document public functions, and move them below, `moshell_exec/init/register/run` are the main method of this file.

vm/tests/flow.rs
syldium, line 1: Who test `assert`?

vm/tests/runner.rs
```diff
@@ -0,0 +1,289 @@
...
+impl<'a> Runner<'a> {
+    pub fn eval(&mut self, expr: &'a str) -> VmValue {
+        match self.try_eval(expr) {
+            Ok(v) => v,
+            Err(VmError::Panic) => panic!("VM did panic"),
+            Err(VmError::Internal) => panic!("VM internal error"),
+        }
+    }
+
+    pub fn try_eval(&mut self, expr: &'a str) -> Result<VmValue, VmError> {
```
syldium, line 70: As those tests are closely related to the VM, you may want to verify that the stack is empty after the execution.

vm/src/vm.h
```diff
@@ -36,6 +70,7 @@ typedef struct moshell_vm {
  * @return an exitcode where:
  *    - 0, the vm exited successfully
  *    - 1, the vm aborted due to a panic
+ *    - -1, tje vm aborted due to an internal error
```
syldium, line 73:
```suggestion
 *    - -1, the vm aborted due to an internal error
```

vm/src/vm.h
```diff
@@ -59,12 +94,18 @@ int moshell_vm_register(moshell_vm vm, const char *bytes, size_t byte_count);
 
 /**
  * Executes the remaining bytecode pages in the VM.
+ * VM runtime does not supports threading and this function must not be called concurrently
```
syldium, line 97:
```suggestion
 *
 * This function is not thread-safe.
```

vm/src/vm.cpp
```diff
@@ -4,13 +4,83 @@
...
+            static_assert(sizeof(T) != sizeof(T), "missing object variant");
+        }
+    },
+               obj->get_data());
+
+    return moshell_object{type, obj};
+}
+
+moshell_value moshell_object_unbox(moshell_object o) {
+    msh::obj *obj = (msh::obj *)o.val;
```
syldium, line 46:
```suggestion
    const msh::obj *obj = static_cast<const msh::obj *>(o.val);
```

vm/src/memory/heap.h
```diff
@@ -20,7 +20,7 @@ namespace msh {
         using std::vector<obj *>::vector;
     };
 
-    using obj_data = std::variant<int64_t, double, const std::string, obj_vector>;
+    using obj_data = std::variant<int64_t, double, std::string, obj_vector>;
```
syldium, line 23:
```suggestion
    using obj_data = std::variant<int64_t, double, const std::string, obj_vector>;
```

vm/src/vm.cpp
```diff
@@ -40,14 +110,23 @@ int moshell_vm_register(moshell_vm vm, const char *bytes, size_t byte_count) {
     return -1;
 }
 
+moshell_value moshell_vm_get_exported(moshell_vm vm, char *name) {
```
syldium, line 113:
```suggestion
moshell_value moshell_vm_get_exported(moshell_vm vm, const char *name) {
```

lib/std/assert.msh
```diff
@@ -0,0 +1,11 @@
+
+
+fun assert(test: Bool) = {
+    if !$test
+        reef::std::panic("assertion failed")
```
syldium, line 5:
```suggestion
    if !$test {
        reef::std::panic("assertion failed")
    }
```

compiler/src/bytecode.rs
```diff
@@ -85,6 +85,11 @@ impl Bytecode {
     }
 }
 
+pub struct InstructionPos {
```
syldium, line 88: Document public types.

vm/tests/runner.rs
```diff
@@ -0,0 +1,289 @@
...
+                None
+            } else {
+                Some(SourceId(current_page - 1))
+            },
+        };
+
+        let mut analysis = self.analyzer.inject(inject, &mut importer, &self.externals);
+        let diagnostics = analysis.take_diagnostics();
+
+        let reef = self.externals.current;
```
syldium, line 93: Read the result from the `Analysis` struct instead of counting pages manually.

vm/tests/flow.rs
```diff
@@ -0,0 +1,140 @@
...
+        Ok(VmValue::Void)
+    )
+}
+
+#[test]
+fn operators() {
+    let mut runner = Runner::default();
+    runner.eval("use std::assert::*");
+    runner.eval(
+        r#"\
```
syldium, line 83:
```suggestion
    runner.eval(
        r#"
        use std::assert::*
```

syldium (changes requested):

lib/std/assert.msh
Override-6: no

vm/binding/lib.rs
```diff
@@ -80,17 +108,113 @@ impl Default for VM {
...
+pub union VmValueFFI {
+    int: i64,
+    byte: u8,
+    double: f64,
+    ptr: *const ffi::c_void,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+enum VmObjectType {
```
syldium, line 133:
```suggestion
#[allow(dead_code)]
enum VmObjectType {
```

vm/binding/lib.rs
```diff
@@ -1,5 +1,15 @@
-use context::source::ContentId;
+#![allow(unused, dead_code)]
+
```
syldium, line 2:
```suggestion
```

vm/binding/lib.rs
```diff
@@ -1,5 +1,15 @@
...
 use std::ffi;
+use std::ffi::{CStr, CString};
+
+use crate::gc::GC;
+use context::source::ContentId;
+
+use crate::value::VmValue;
```
syldium, line 9:
```suggestion
```

vm/binding/value.rs
```diff
@@ -0,0 +1,42 @@
+use crate::{VmObjectFFI, VmObjectType, VmValueFFI};
```
syldium, line 1:
```suggestion
use crate::{VmObjectFFI, VmObjectType};
```

vm/binding/lib.rs
```diff
@@ -80,17 +108,113 @@ impl Default for VM {
...
+    Vec,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct VmObjectFFI(VmObjectType, *mut ffi::c_void);
+
+impl VmValueFFI {
+    /// # Safety
+    /// safety contract must be held by caller
```
syldium, line 146:
```suggestion
    /// The caller must ensure that the value has the correct type.
```

vm/tests/runner.rs
```diff
@@ -0,0 +1,283 @@
...
+use context::source::Source;
+use parser::parse_trusted;
+use vm::value::VmValue;
+use vm::{VmError, VmValueFFI, VM};
+
+pub struct Runner<'a> {
+    externals: Externals<'a>,
+    vm: VM,
+    analyzer: Analyzer<'a>,
+    current_page: Option<usize>,
```
syldium, line 21:
```suggestion
    current_page: Option<SourceId>,
```

vm/tests/runner.rs
```diff
@@ -0,0 +1,283 @@
...
+        let mut bytes = Vec::new();
+
+        let chunk = self.analyzer.engine.get_user(page).unwrap();
+        let chunk_expr = &chunk.expression.as_ref().unwrap();
+
+        let evaluated_expr_type = if let ExprKind::Block(exprs) = &chunk_expr.kind {
+            exprs.last().map_or(types::UNIT, |v| v.ty)
+        } else {
+            chunk_expr.ty
+        };
```
syldium, line 105:
```suggestion
        let evaluated_expr_type = chunk_expr.ty;
```
The typing phase already does that, recursively.

vm/src/vm.h
```diff
@@ -80,6 +83,114 @@ size_t moshell_vm_next_page(moshell_vm vm);
...
 
+/**
+ * A moshell value, either an i64, a unsigned byte, a double
+ * or an address
+ * */
+typedef union {
+    int64_t i;
+    uint8_t b;
+    double d;
+    const void *ptr;
```
syldium, line 94:
```suggestion
    void const *ptr;
```
The pointer is `const`, not the value behind.

syldium (changes requested):

vm/src/vm.h
Override-6: both

vm/binding/lib.rs
```diff
@@ -67,7 +84,14 @@ impl VM {
...
     /// Gets the next page of bytecode to be executed.
     pub fn get_next_page(&self) -> ContentId {
-        ContentId(unsafe { moshell_vm_next_page(self.0) })
+        ContentId(unsafe { moshell_vm_next_page(self.ffi) })
+    }
+
+    pub fn get_exported_var(&self, name: &str) -> VmValueFFI {
+        unsafe {
+            let c_name = CString::new(name).unwrap();
+            moshell_vm_get_exported(self.ffi, c_name.into_raw())
```
syldium, line 93:
```suggestion
            moshell_vm_get_exported(self.ffi, c_name.as_ptr())
```
The string is still owned by this function (or even better, don't take a C-string in the FFI by using a pointer and a size).

vm/binding/lib.rs
```diff
@@ -80,17 +104,114 @@ impl Default for VM {
...
 #[repr(C)]
 #[derive(Copy, Clone)]
-struct VmFFI(*mut ffi::c_void);
+struct VmArrayFFI(u64, *mut VmValueFFI);
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct VmStringFFI(u64, *mut ffi::c_char);
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub union VmValueFFI {
```
syldium, line 120: Keep this struct internal and promote the use of the already present safe alternative would be great.

If unnecessary conversions between the two interface are an issue, maybe treat the Rust part a simple safe wrapper with lifetimes?

vm/src/vm.h
```diff
@@ -80,6 +83,114 @@ size_t moshell_vm_next_page(moshell_vm vm);
...
+typedef struct {
+    moshell_object_type type;
+    const void *val;
+} moshell_object;
+
+/**
+ * A sized array of values
+ * */
+typedef struct {
+    const uint64_t size;
```
syldium, line 119:
```suggestion
    size_t size;
```
I would drop the `const` qualifier here, as the view does not impact the actual vector, and to be in sync with the other structs that don't have it. Anyway, the main thing here is that as the vector implementation depend on the host (the size cannot be greater than the size of a pointer), using platform dependent types here make sense.

syldium (changes requested):
Cool stuff! Address the memory management issues/unsafe things, so it can be merged.

vm/src/vm.h
Override-6: it feels wrong that the array's size can change but not its pointer, i added more consts to other structs to be more in sync

vm/binding/lib.rs
Override-6: You need to know the kind of value held by the VmValuFFI transform it to a safe VmValue, however the VM can't know what's behind a VmValueFFI si it can only return an unsafe holder to a value. It's up to the user to use its convert methods safely

syldium: Exposing an unsafe type and a safe type together without conversion functions, even unsafe ones, is bizarre. The point here is that this would require exposing the `analyzer` crate at the `dependencies` level, which is effectively meh.

vm/binding/lib.rs
```diff
@@ -26,14 +40,19 @@ pub enum VmError {
     Internal,
 }
 
+#[allow(dead_code)]
```
syldium, line 43:
```suggestion
```

syldium (approved):

PR #162, opened 2023-10-12
Fix bounds handling for methods + Vec::pop patch

Override-6:
This PR fixes previous hazardous behavior when applying type bounds to methods with parameters typed with type's generics.
This allowed the `Vec::pop` method to return an Option[A]


syldium (approved):
Please reformat your code. Otherwise, it looks very good.

PR #164, opened 2023-10-14
Add sh-like program arguments access

Override-6:
this pull request adds the magic variables `$n`, `$#`, `$@` and `$*`.

A reference to those variables will init a new global string vector that contains the program arguments.
Thus, type of `$@` and `$*` is a `Vec[String]`.  
To shift arguments, you'll have to call the new Vec::pop_head method on the variable : `$@.pop_head()`.

Program arguments is passed to the VM through the `moshell_vm_init` FFI


analyzer/src/lib.rs
```diff
@@ -256,3 +257,7 @@ pub struct ResolutionResult<'e> {
     imports: Imports,
     visited: HashSet<Name>,
 }
+
+pub fn is_magic_variable_name(name: &str) -> bool {
```
syldium, line 261:
```suggestion
/// Checks if the given name is reserved for an external variable that is not
/// defined in the source.
pub fn is_magic_variable_name(name: &str) -> bool {
```

syldium (changes requested):

vm/src/stdlib_natives.cpp
```diff
@@ -3,6 +3,8 @@
 #include "memory/heap.h"
 #include <charconv>
 #include <cmath>
+#include <cstring>
+#include <fstream>
```
syldium, line 7:
```suggestion
```

cli/src/cli.rs
```diff
@@ -42,6 +41,9 @@ pub struct Cli {
     /// Generate tab-completion scripts for your shell
     #[arg(long = "completions")]
     pub(crate) completions: Option<Shell>,
+
+    #[arg(last = true)]
```
syldium, line 45:
```suggestion
    /// The arguments to pass to the script
    #[arg(index = 2)]
```

cli/src/cli.rs
```diff
@@ -24,7 +24,6 @@ use crate::report::{display_diagnostic, display_parse_error};
 #[command(author, version, about, long_about = None)]
 pub struct Cli {
     /// Defines the source file to parse
-    #[arg(index = 1)]
```
syldium, line 27:
```rs
    #[arg(index = 1)]
```

syldium (changes requested):

syldium (approved):

Issue #166, opened 2023-10-20
Wrap every C++ code in a namespace


syldium:
While some code is in the `msh` namespace, some older code isn't. It would be great to uniformize that. This would also drastically reduce the chance of a function name clash, with the `nltoh` function on MacOS for instance .

PR #168, opened 2023-10-21
Add macOS support

syldium:
The nightly setup script now checks the operating system and the architecture. Binaries are now built for the Darwin target. It fixes the macro conflict reported in #166.

The binary for macOS is crafted with the system Apple Clang compiler, that doesn't support C++ 20 concepts, so they are removed before the build. It may be better to create a C macro that checks for the compiler.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,6 +1,19 @@
+#!/bin/sh
+# shellcheck shell=dash
+
 set -eu
 
-NIGHTLY_URL=https://moshell.dev/releases/nightly
+if ! [ "$(uname -m)" = "x86_64" ]; then
+  echo "Unsupported $(uname -m) architecture" >&2
+  exit 1
+fi
+case "$(uname -s)" in
+  Linux) target=x86_64-linux-gnu;;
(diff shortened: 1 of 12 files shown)


setup.sh
```diff
@@ -1,6 +1,19 @@
...
+  echo "Unsupported $(uname -m) architecture" >&2
+  exit 1
+fi
+case "$(uname -s)" in
+  Linux) target=x86_64-linux-gnu;;
+  Darwin) target=x86_64-apple-darwin;;
+  *) echo "Unsupported $(uname -s) OS" >&2; exit 1;;
+esac
+
+NIGHTLY_URL=https://moshell.dev/releases/nightly/$target
```
Override-6, line 16:
```suggestion
  Linux) TARGET=x86_64-linux-gnu;;
  Darwin) TARGET=x86_64-apple-darwin;;
  *) echo "Unsupported $(uname -s) OS" >&2; exit 1;;
esac

NIGHTLY_URL=https://moshell.dev/releases/nightly/$TARGET
```

Override-6 (approved):

PR #175, opened 2023-11-21
Implement background processes

### LINKED ISSUE #174: support detach suffix
implement the detach suffix operator `&` on analyzer and virtual machine.

syldium:
Type subprocesses (`(cmd)`) and background processes (`cmd &`).

Because a forked process should always be awaited, the `std::process` module has been added.

- Fixes #174

### DIFF
--- a/analyzer/src/steps/typing.rs
+++ b/analyzer/src/steps/typing.rs
@@ -44,7 +44,7 @@ use crate::types::operator::name_operator_method;
 
 use crate::types::ty::{FunctionType, Type, TypeRef};
 use crate::types::{
-    builtin, Typing, BOOL, ERROR, EXITCODE, FLOAT, GLOB, INT, NOTHING, STRING, UNIT, PID
+    builtin, Typing, BOOL, ERROR, EXITCODE, FLOAT, GLOB, INT, NOTHING, PID, STRING, UNIT,
 };
 
 mod assign;
--- a/analyzer/src/types/builtin.rs
+++ b/analyzer/src/types/builtin.rs
@@ -8,7 +8,10 @@ use crate::types::ctx::TypeContext;
 use crate::types::engine::TypedEngine;
 use crate::types::operator::name_operator_method;
 use crate::types::ty::{MethodType, Type, TypeId, TypeRef};
-use crate::types::{Typing, BOOL, ERROR, EXITCODE, FLOAT, GENERIC_OPTION, GENERIC_VECTOR, GLOB, INT, NOTHING, STRING, UNIT, PID};
+use crate::types::{
+    Typing, BOOL, ERROR, EXITCODE, FLOAT, GENERIC_OPTION, GENERIC_VECTOR, GLOB, INT, NOTHING, PID,
+    STRING, UNIT,
+};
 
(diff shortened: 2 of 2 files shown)


analyzer/src/steps/typing.rs
```diff
@@ -777,6 +777,64 @@ fn ascribe_substitution(
...
+        diagnostics,
+        &detached.underlying,
+        state.with_local_value(ExpressionValue::Unused),
+    );
+    TypedExpr {
+        kind: ExprKind::Subprocess(Subprocess {
+            inner: Box::new(expr),
+            awaited: false,
+        }),
+        ty: INT,
```
Override-6, line 799: why not introduce a PID type ?

analyzer/src/types/hir.rs
```diff
@@ -74,6 +74,12 @@ pub struct MethodCall {
     pub function_id: FunctionId,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct Subprocess {
+    pub inner: Box<TypedExpr>,
+    pub awaited: bool,
```
Override-6, line 80:
```suggestion
    pub detached: bool,
```
Then invert the bools

vm/src/stdlib_natives.cpp
```diff
@@ -310,6 +311,20 @@ static void program_arguments(OperandStack &caller_stack, runtime_memory &mem) {
...
+    int status;
+    if (waitpid(pid, &status, 0) == -1) {
+        throw RuntimeException("Failed to wait for process " + std::to_string(pid) + ": " + strerror(errno) + ".");
+    }
+}
+
+static void process_wait_all(OperandStack &, runtime_memory &) {
+    int status;
+    while (wait(&status) > 0)
+        ;
```
Override-6, line 325:
```suggestion
    while (wait(&status) > 0);
```

Override-6 (approved):

vm/src/stdlib_natives.cpp
syldium: Are you referring to `// clang-format off`? This is the format for `while` statements that don't have bodies.

analyzer/src/types/hir.rs
syldium: All those processes are "detached" from the main one: they don't share the same execution thread. The fact that some are awaited and some aren't is the only difference here. Also, I prefer using a field that isn't inverted later not to add unnecessary code bloat.

analyzer/src/steps/typing.rs
syldium: I have mixed feelings about this. On one hand, this would be useful to avoid misuses and on the other hand, the current native system isn't ready to nicely integrate type aliases/native wrapper types.

vm/src/stdlib_natives.cpp
Override-6: its disgusting (not opinionated)

syldium: To kill two birds with one stone, a *clang-tidy* check may also be suppressed here by declaring an empty body. *clang-format* creates:
```c
    while (wait(&status) > 0) {
    }
```

PR #178, opened 2023-12-21
Implement for in iterables

syldium:
Adds a simpler way to iterate over vectors, integer ranges and strings.

All those iteration types are hard-coded with internal int indexes. Indexes over a vector are always incremented by one, whereas indexes in strings use the length in bytes of the current char to account for multi bytes characters.

Strings are byte-indexed, and may not be indexed in the middle of a multibyte character.

Int ranges are describes as structures in the standard library, with a variant for inclusive ranges.


Override-6 (approved):
very cool 👍🏻

syldium:
This PR is currently pending a fix against stack pollution in range iterations.
Example:
```
echo $((90 + { for n in 1..3 {}; 11 }))
```

PR #181, opened 2024-06-12
Fuzz the parser

syldium:
Use LLVM's libFuzzer to detect inputs that cause a panic or an infinite loop. Those that are quickly detectable with this setup should already be fixed.

The fuzzer can generate unnatural inputs, which are still valid UTF-8. Because every unknown character used to be considered an identifier, this led to bizarre situations. The identifier alphabet is now restricted to accept only Unicode identifiers, ASCII digits and the underscore character.

A nightly toolchain must be used to run it, and [`cargo fuzz`](https://github.com/rust-fuzz/cargo-fuzz) should be used:
```sh
cargo +nightly fuzz run parse -- -timeout=15 -max_len=600
```


Override-6 (approved):
Why not adding it to the CI ?

syldium:
> Why not adding it to the CI ?

The fuzzer tries random bit patterns indefinitely, and doesn't really stop until there's a timeout or a signal. The result it gives also depends heavily on the initial corpus, and may not be related to the commit just pushed.

